PR #12, opened 2016-07-19
Uob 43 add promise

gvdent:


gvdent:
@malamili

malamili:
adds() is a bit confusing, or is this from some sort of convention? Would addPromise not be beter, then have function(entity, error) resolve in promise?

malamili:
does the promise btw correctly reject on server error?

src/weaver.coffee
```diff
@@ -54,10 +55,21 @@ class Weaver
     attributes[key] = value for key, value of data when not isEntity(value)
     relations[key]  = value.$id() for key, value of data when isEntity(value)
 
-    @channel.create({type, id:entity.$id(), attributes, relations})
+    createPromise = @channel.create({type, id:entity.$id(), attributes, relations})
```
malamili: does this correctly rejects?

gvdent:
Added a test to see whether reject is called on disconnect. So if the server correctly rejects when an error is called, the promise returned by addPromise should reject as well.

PR #16, opened 2016-09-07
Entity not found

mingushead:
Listens for exceptions thrown from weaver-server.

### DIFF
--- a/src/socket.coffee
+++ b/src/socket.coffee
@@ -2,6 +2,10 @@
 io      = require('socket.io-client')
 Promise = require('bluebird')
 
+isError = (object) ->
+
+  object? and Object.keys(object).length is 3 and object.code? and object.payload? and object.message?
+
 # Transport layer to server using socket.io
 module.exports =
 class Socket
@@ -53,10 +57,6 @@ class Socket
 
   # TODO: Handle errors from server
   emit: (key, body) ->
-
-    isError = (object) ->
-      return false if not object?
-      Object.keys(object).length is 3 and object.code? and object.payload? and object.message?
 
     new Promise((resolve, reject) =>
       @io.emit(key, body, (response) ->
--- a/src/socket.coffee
+++ b/src/socket.coffee
@@ -3,7 +3,6 @@ io      = require('socket.io-client')
 Promise = require('bluebird')
 
 isError = (object) ->
-
   object? and Object.keys(object).length is 3 and object.code? and object.payload? and object.message?
 
 # Transport layer to server using socket.io
@@ -61,7 +60,9 @@ class Socket
     new Promise((resolve, reject) =>
       @io.emit(key, body, (response) ->
         if isError(response)
-          reject(new Error(response.code + ' - ' + response.message))
+          error = response
+          error.isError = -> true
+          resolve(error)
         else
           resolve(response)
       )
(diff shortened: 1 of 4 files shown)


src/socket.coffee
```diff
@@ -53,10 +53,15 @@ class Socket
 
   # TODO: Handle errors from server
   emit: (key, body) ->
+
+    isError = (object) ->
```
malamili: define this function in top of file in line 4 to save resources (not evaluate this on every emit)

src/socket.coffee
```diff
@@ -53,10 +53,15 @@ class Socket
...
+      return false if not object?
```
malamili: just add 'object?' in the second check:

`
object? and Object.keys(object).length is 3 and object.code? and object.payload? and object.message?
`

src/socket.coffee
```diff
@@ -53,10 +53,15 @@ class Socket
...
+      Object.keys(object).length is 3 and object.code? and object.payload? and object.message?
+
     new Promise((resolve, reject) =>
       @io.emit(key, body, (response) ->
-        if response is 0
-          resolve()
+        if isError(response)
+          reject(new Error(response.message))
```
malamili: what are we doing with code? Shouldnt we do new Error(response)?

src/weaver.coffee
```diff
@@ -84,30 +84,29 @@ class Weaver
   # Core
   # Loads an entity either from the local repository or from the server
   get: (id, opts) ->
-    # Default options
-    opts = {} if not opts?
-    opts.eagerness = 1 if not opts.eagerness?
+      # Default options
```
malamili: why are these changes? according to .editorconfig in the root file a tab is 2 spaces. is some other editor adding spaces?

test/weaver/weaver.test.coffee
```diff
@@ -121,20 +121,23 @@ describe 'Weaver: addPromise', ->
 describe 'Weaver: Creating an entity', ->
 
   it 'should set type as $ROOT if left empty', ->
-    return
+    root = weaver.add()
```
malamili: very nice!

malamili:
Thnx for testing! 👍

src/socket.coffee
mingushead: Well, the response looks like this:
response =
     code: 404
     message: 'Entity not found'
     payload: {type:"some-type"..}

Which means the exception as seen from the browser looks like this:

screen shot 2016-09-08 at 11 06 31

rejecting with message property of the response however, means the exception will look like this in the browser:

screen shot 2016-09-08 at 11 10 55

mingushead: Perhaps I should change the structure of the response coming from the weaver-server? Let me know what you think..

src/socket.coffee
```diff
@@ -53,10 +57,11 @@ class Socket
 
   # TODO: Handle errors from server
   emit: (key, body) ->
+
     new Promise((resolve, reject) =>
       @io.emit(key, body, (response) ->
-        if response is 0
-          resolve()
+        if isError(response)
+          reject(new Error(response.code + ' - ' + response.message))
```
malamili: I think its better to reject with just the response:

reject(response)

PR #20, opened 2017-01-11
Node ops

malamili:


src/WeaverNode.coffee
```diff
@@ -1,78 +1,139 @@
...
 
 
-  destroy: ->
+  # Clear all pendingwrites, used for instance after saving or when loading a node
+  clearPendingWrites: ->
+    @pendingWrites = []
+
+    for key, relation of @relations
+      for id, node of relation.nodes
+        node.clearPendingWrites()
```
gvdent: - This recursively clears all pending writes on the in memory structure, not just for this single node. The name and comment of the function do not reflect that destructive nature.
- Should the in memory structure contain any loops through relations back to this node, this is an endless loop and the end of the running application.

src/WeaverNode.coffee
```diff
@@ -1,78 +1,139 @@
...
+  collectPendingWrites: ->
+    operations = @pendingWrites
 
-    # Check to create the node first
-    if not @saved
-      operations.push(new Operation.Node(@).create())
-      @saved = true
+    for key, relation of @relations
+      for id, node of relation.nodes
+        operations = operations.concat(node.collectPendingWrites())
```
gvdent: Possible endless loop

src/WeaverNode.coffee
```diff
@@ -1,78 +1,139 @@
...
 
-  fetch: ->
+
+  # Checks wether needs saving
+  isDirty: ->
+    @pendingWrites isnt []
+
+
+  # Save node and all values / relations and relation objects to server
+  save: (values) ->
```
gvdent: Why don't we store pending operations with the coremanager for instance so we don't have to do all the recursing and looping over nodes and relations when saving? Is there an advantage to storing these pending changes all over the place instead of in a central location?

src/WeaverNode.coffee
malamili: haha good point

src/WeaverNode.coffee
malamili: verrrrryyy good point, thnx, thats a much better approach..blame it on feature evolution that caused this mess haha

PR #21, opened 2017-01-12
Auth1

carvilsi:
adding permission call and test


test/WeaverUser.test.coffee
```diff
@@ -7,7 +7,7 @@ require('./../src/WeaverNode')  # This preloading will be an issue
 require('./../src/WeaverUser')
 
 describe 'Weaver User', ->
-  this.timeout(2000)
+  this.timeout(2500)
```
gvdent: Is the service so slow that this timeout is actually required?

carvilsi: Really does not… I'm reverting it to 2000

src/WeaverUser.coffee
```diff
@@ -22,6 +22,17 @@ class WeaverUser
       catch error
       res
     )
+    
+  permission: (usr) ->
+    coreManager = Weaver.getCoreManager()
+    users = Weaver.getUsersDB()
+    @current(usr).then((token) ->
+      userPayload = {
```
malamili: this can be rewritten as userPayload = {user, acces_token} like mentioned in previous PR (change parameters usr to user and token to access_token)

carvilsi: Yep :)

malamili:
Thnx for that change. Would you also please change access_token to accessToken?

carvilsi:
I'm going to change to accessToken before the merge

PR #22, opened 2017-01-13
Projects

gvdent:
Calling weaver-server projects functionality. Current supports create/delete/list, actually using these will be a seperate PR.


src/WeaverProject.coffee
```diff
@@ -1,27 +1,23 @@
...
+  create: ->
+    Weaver.getCoreManager().createProject(@).then((res) =>
+      @id = JSON.parse(res).id
+      @
+    )
 
+  delete: ->
+    Weaver.getCoreManager().deleteProject(@)
 
+  @list: ->
```
malamili: I like that this doesnt cache anything

PR #23, opened 2017-01-16
Auth2

carvilsi:
messaging for errors

### DIFF
--- a/test/WeaverUser.test.coffee
+++ b/test/WeaverUser.test.coffee
@@ -122,7 +122,9 @@ describe 'Weaver User', ->
         user.signUp('phoenix','andromeda','andromeda@univer.se','andromedas','SYSUNITE')
       ).then(->
         user.signUp('phoenix','andromeda','centaurus@univer.se','andromedas','SYSUNITE')
-      ).then().catch((error)->
+      ).then(->
+        assert(false)
+      ).catch((error)->
         assert.equal(error.code, WeaverError.DUPLICATE_VALUE)
       )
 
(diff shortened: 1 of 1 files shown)


test/WeaverUser.test.coffee
```diff
@@ -114,3 +114,25 @@ describe 'Weaver User', ->
...
+    
+  it 'should fails trying to signUp with an existing userName', ->
+      user = new Weaver.User()
+      user.logIn('phoenix','phoenix')
+      .then(->
+        user.signUp('phoenix','andromeda','andromeda@univer.se','andromedas','SYSUNITE')
+      ).then(->
+        user.signUp('phoenix','andromeda','centaurus@univer.se','andromedas','SYSUNITE')
+      ).then().catch((error)->
+        assert.equal(error.code, WeaverError.DUPLICATE_VALUE)
```
gvdent: This test only asserts that the specific error code is passed iff an error is given.

maybe add a ....then( ->
  assert(false)
) at the end so the test fails if no error is given.

test/WeaverUser.test.coffee
```diff
@@ -114,3 +114,25 @@ describe 'Weaver User', ->
...
+      )
+
+    
+  it 'should fails trying to signUp with an existing userEmail', ->
+      user = new Weaver.User()
+      user.signUp('phoenix','andromedas','andromedas@univer.se','andromedas','SYSUNITE')
+      .then(->
+        user.signUp('phoenix','andro','andromedas@univer.se','andromedas','SYSUNITE')
+      ).then().catch((error)->
+        assert.equal(error.code, WeaverError.DUPLICATE_VALUE)
```
gvdent: Same as previous comment

test/WeaverUser.test.coffee
carvilsi: 👍

test/WeaverUser.test.coffee
carvilsi: 👍

PR #24, opened 2017-01-16
making logIn, logOut, signUp, signOff static

carvilsi:
logIn, logOut, signUp, signOff static
The requirement was to change just logOut and signUp, but logOut and signOff was added too, trying to be more consistent

### DIFF
--- a/src/WeaverUser.coffee
+++ b/src/WeaverUser.coffee
@@ -4,14 +4,14 @@ Weaver = require('./Weaver')
 loki   = require('lokijs')
 Error        = require('weaver-commons').Error
 WeaverError  = require('weaver-commons').WeaverError
+WeaverNode   = require('./WeaverNode')
 
-class WeaverUser
+class WeaverUser extends WeaverNode
   
   constructor: (@email, @username, @password) ->
     @emailVerified = false
     
-    
-  logIn: (user, password) ->
+  @logIn: (user, password) ->
     credentials = {user,password}
     coreManager = Weaver.getCoreManager()
     users = Weaver.getUsersDB()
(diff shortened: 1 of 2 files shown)


src/WeaverUser.coffee
```diff
@@ -25,24 +25,26 @@ class WeaverUser
...
-    users = Weaver.getUsersDB()
-    @current(user).then((accessToken) ->
+    weaverUser = new WeaverUser()
+    weaverUser.current(user).then((accessToken) ->
       userPayload = {user,accessToken}
       coreManager.permission(userPayload)
     )
   
-  signUp: (currentUsr,userName,userEmail,userPassword,directoryName) ->
+  @signUp: (currentUsr,userName,userEmail,userPassword,directoryName) ->
```
malamili: What is the role of currentUsr here?

test/WeaverUser.test.coffee
```diff
@@ -101,42 +96,36 @@ describe 'Weaver User', ->
...
-
-    
+  
   it 'should fails trying to signUp with an existing userEmail', ->
-      user = new Weaver.User()
-      user.signUp('phoenix','andromedas','andromedas@univer.se','andromedas','SYSUNITE')
+      Weaver.User.signUp('phoenix','andromedas','andromedas@univer.se','andromedas','SYSUNITE')
       .then(->
-        user.signUp('phoenix','andro','andromedas@univer.se','andromedas','SYSUNITE')
+        Weaver.User.signUp('phoenix','andro','andromedas@univer.se','andromedas','SYSUNITE')
```
malamili: What is phoenix?

src/WeaverUser.coffee
malamili: I dont understand why you need currentUser for signup, signoff, etc.

src/WeaverUser.coffee
```diff
@@ -63,7 +65,7 @@ class WeaverUser
   # passing a user is optional
   ###
   
-  logOut: (usr) ->
+  @logOut: (usr) ->
```
malamili: logOut is strange to be static..use static only when there isnt an object. when logging in, you have a user object that u can logout

src/WeaverUser.coffee
carvilsi: We do not have any way to signUp but with a user that already exits at weaver-flock. This current user is the user that is loggedin and has an accessToken, and can perform the signUp process.

test/WeaverUser.test.coffee
carvilsi: phoenix is the default user at weaver-flock

src/WeaverUser.coffee
carvilsi: Ok, I'will change this to non static

PR #25, opened 2017-01-16
Auth4

carvilsi:
adding messaging for logIn case that user or password are not right. 
Return to receive array for the permission list of user instead a string

### DIFF
--- a/src/WeaverUser.coffee
+++ b/src/WeaverUser.coffee
@@ -4,14 +4,14 @@ Weaver = require('./Weaver')
 loki   = require('lokijs')
 Error        = require('weaver-commons').Error
 WeaverError  = require('weaver-commons').WeaverError
+WeaverNode   = require('./WeaverNode')
 
-class WeaverUser
+class WeaverUser extends WeaverNode
   
   constructor: (@email, @username, @password) ->
     @emailVerified = false
     
-    
-  logIn: (user, password) ->
+  @logIn: (user, password) ->
     credentials = {user,password}
     coreManager = Weaver.getCoreManager()
     users = Weaver.getUsersDB()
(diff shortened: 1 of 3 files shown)


src/WeaverUser.coffee
```diff
@@ -25,24 +25,26 @@ class WeaverUser
...
       newUserPayload = {newUserCredentials,accessToken}
       coreManager.signUp(newUserPayload)
     )
     
   
   signOff: (currentUsr, user) ->
     coreManager = Weaver.getCoreManager()
-    @current(currentUsr).then((accessToken) ->
+    weaverUser = new WeaverUser()
+    weaverUser.current(currentUsr).then((accessToken) ->
```
gvdent: Can signOff now be static (@signOff) as well?

test/WeaverUser.test.coffee
```diff
@@ -19,80 +22,94 @@ describe 'Weaver User', ->
...
+    weaverUser = new Weaver.User()
+    weaverUser.permission('phoenix').then((res) ->
+      expect(res).to.eql(['read_user', 'create_user', 'delete_user', 'create_role', 'read_role', 'delete_role', 'create_permission', 'read_permission', 'delete_permission', 'read_application', 'create_application', 'delete_application', 'create_directory', 'read_directory', 'delete_directory'])
     )
-    
+  
   it 'should fails when trying to login with non existing user', ->
-    user = new Weaver.User()
-    user.logIn('andromeda','Chains')
+    Weaver.User.logIn(randomstring.generate({length:5,charset:'alphabetic'}),randomstring.generate(7))
```
gvdent: In general: refrain from using random anything in unit tests, it damages reproducability. A test containing random may fail or pass depending on the random value, resulting in a horrible heisentestfailure.

src/WeaverUser.coffee
carvilsi: 👍

test/WeaverUser.test.coffee
carvilsi: I'm going to change but for the test environment this should fails always.

PR #27, opened 2017-01-17
adding creating a application to SDK. Making some old error messages …

carvilsi:
Adding creating a application to SDK. 
Making some old error messages nicer. 
Also sending an error message instead null when there is no jwt for the current(user)


test/WeaverApplication.test.coffee
```diff
@@ -0,0 +1,53 @@
...
+    weaverApplication = new Weaver.Application()
+    Weaver.User.logIn('phoenix','phoenix')
+    .then( ->
+      weaverApplication.createApplication('phoenix','fooApplication1','barProject1')
+    ).then((res) ->
+      assert.equal(res.success,'Document created')
+    )
+    
+
+  it 'should fails on creates an application', ->
```
gvdent: The test description should also specify why it fails (e.g. 'should fail to create an application without a user session'), this becomes especially useful when there are multiple reasons why a creation can fail.

src/WeaverApplication.coffee
```diff
@@ -0,0 +1,27 @@
...
+    @weaverUser = new Weaver.User()
+    
+  createApplication: (user,applicationName,projectName) ->
+    coreManager = Weaver.getCoreManager()
+    @weaverUser.current(user)
+    .then((accessToken) ->
+      newApplication = {projectName,applicationName,accessToken}
+      coreManager.createApplication(newApplication)
+    ).catch((err) ->
+      Promise.reject(err)
```
gvdent: You can leave this catch out, the function will already return a Promise. Without the catch that one will be rejected with err as the argument, making this statement redundant.

carvilsi: 👍

PR #28, opened 2017-01-17
Projects activate

malamili:


src/WeaverNode.coffee
```diff
@@ -122,18 +122,18 @@ class WeaverNode
 
 
   # Save node and all values / relations and relation objects to server
-  save: (values) ->
+  save: (project) ->
```
gvdent: More of a design philosophy thing: Do we ever want to support working on one project and meanwhile performing operations on another project? Seems like an edge case which increases complexity of the code with little yield to me, but I may not have the full picture.

malamili: I had to do this because a project is now a node itself which saves to a project..so imagine everytime somebody is changing the project name, we need to do current = Weaver.getCurrent(); Weaver.useProject(projectToChange). do the change..then Weaver.useProjecT(current) again..and this looked like a mess to me..but perhaps there is a better way?

src/WeaverProject.coffee
```diff
@@ -1,21 +1,53 @@
...
+            else
+              resolve()
+          )
+
+        checkReady()
+      )
+    )
+    .then(=> # Project is ready, create the node
+      @_created = true
+      @save()
```
malamili: for instance here, I would had to change currentProject otherwise first and change it back

src/WeaverNode.coffee
gvdent: Ok... good reason :) We just have to be extra careful that this parameter is only ever required for these edge cases (we don't want app developers fingers to get tired from passing the project all the time when not required)

malamili: indeed, its optional :)

PR #32, opened 2017-01-26
File0

carvilsi:
Uploading and downloading for sdk
UUID is added as ID with each file and the user can request the file by ID or by complete filename (UUID-fileName.ext)

### DIFF
--- a/src/WeaverFile.coffee
+++ b/src/WeaverFile.coffee
@@ -57,9 +57,9 @@ class WeaverFile extends Weaver.SystemNode
     coreManager.getFileByID(file)
     .then((buffer) ->
       if buffer.code?
-        Promise.reject(Error WeaverError.FILE_NOT_EXISTS_ERROR,"The requested file #{fileName} at #{project} does not exits")
+        Promise.reject(Error WeaverError.FILE_NOT_EXISTS_ERROR,"The requested file #{id} at #{project} does not exits")
       else if Object.keys(buffer).length is 0
-        Promise.reject(Error WeaverError.FILE_NOT_EXISTS_ERROR,"The requested file #{fileName} can\'t be retrieved because #{project} does not exists")
+        Promise.reject(Error WeaverError.FILE_NOT_EXISTS_ERROR,"The requested file #{id} can\'t be retrieved because #{project} does not exists")
       else
         writeFile(path, buffer)
     ).catch((err) ->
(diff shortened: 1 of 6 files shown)


src/WeaverFile.coffee
```diff
@@ -7,6 +12,58 @@ class WeaverFile extends Weaver.SystemNode
...
+    }
+    coreManager.getFile(file)
+    .then((buffer) ->
+      if buffer.code?
+        Promise.reject(Error WeaverError.FILE_NOT_EXISTS_ERROR,"The requested file #{fileName} at #{project} does not exits")
+      else if Object.keys(buffer).length is 0
+        Promise.reject(Error WeaverError.FILE_NOT_EXISTS_ERROR,"The requested file #{fileName} can\'t be retrieved because #{project} does not exists")
+      else
+        writeFile(path, buffer)
+    ).catch((err) ->
```
gvdent: This catch can be removed

malamili:
Perhaps i didnt see it, but is there a test where you a) save a file, b) reload from database and c) actually read the contents of the file to assure it saved correctly?

src/WeaverFile.coffee
```diff
@@ -7,6 +12,58 @@ class WeaverFile extends Weaver.SystemNode
...
+    .then((buffer) ->
+      if buffer.code?
+        Promise.reject(Error WeaverError.FILE_NOT_EXISTS_ERROR,"The requested file #{id} at #{project} does not exits")
+      else if Object.keys(buffer).length is 0
+        Promise.reject(Error WeaverError.FILE_NOT_EXISTS_ERROR,"The requested file #{id} can\'t be retrieved because #{project} does not exists")
+      else
+        writeFile(path, buffer)
+    ).catch((err) ->
+      Promise.reject(err)
+    )
```
gvdent: This catch can be removed. Basically

```
somePromiseReturningMethod().then((res) ->
  Promise.resolve(res)
).catch((err) ->
  Promise.reject(err)
)
``` 
is equivalent to 
```
somePromiseReturningMethod()
```
Since the function called already returns a Promise that is either resolved or rejected with the result. So unless you need to process or want to log the result/error of the first call, just return the Promise that function creates.

src/WeaverFile.coffee
carvilsi: 👍

src/WeaverFile.coffee
carvilsi: 👍

carvilsi:
> Perhaps i didnt see it, but is there a test where you a) save a file, b) reload from database and c) actually read the contents of the file to assure it saved correctly?

I did not perform this test, I've been using a png and checking if it is correct:
I will write it:
a) save a file
b) retrieve it from object storage server
c) read the content and compare with the original one

PR #33, opened 2017-01-27
Projects activate

gvdent:
Actually just a quick fix to the tests for k8s compatibility

### DIFF
--- a/test/WeaverProject.test.coffee
+++ b/test/WeaverProject.test.coffee
@@ -71,7 +71,6 @@ describe 'WeaverProject Test', ->
 
     ).then(->
       done()
-    ).catch((Err) -> console.log Err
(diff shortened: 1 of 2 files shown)


test/WeaverProject.test.coffee
```diff
@@ -69,10 +69,12 @@ describe 'WeaverProject Test', ->
       expect(loadedA.get("name")).to.equal('A')
       expect(loadedB.get("name")).to.be.undefined
 
-      Promise.all([a.destroy(), b.destroy()])
     ).then(->
       done()
-    ).catch((Err) -> console.log Err)
+    ).catch((Err) -> console.log Err
```
carvilsi: Do not know if our BDFL will be agree with this console.log

PR #35, opened 2017-01-31
Introduce model concept

mingushead:
First bit of WeaverModel functionality.

##### So far, allows you to:
- define a model (blueprint of a set of linked individual/value properties with a single parent)
- create instances of that model

##### Caveats:
- based on and depends on chirql
- must be tree-like, ie all parent to child relationships must be one-to-one, or one-to-many, but never many-to-one/many-to-many
-  lots of features yet to be implemented, too many to list here; the tests showcase all of the current capabilities.


src/WeaverModel.coffee
```diff
@@ -0,0 +1,207 @@
...
+
+
+class WeaverModel
+
+  constructor: (@name)->
+
+    @lexer = new Lexer()
+    @parser = new Parser()
+
+    @instance = =>
```
malamili: why define these functions in the constructor instead of:
```instance: -> ```

File: src/WeaverModel.coffee [hunk 2]
```diff
@@ -0,0 +1,207 @@
...
+      tokens = @lexer.lex(@definitionString)
+      fragmentList = @parser.parseTokens(tokens)
+      @definition = {}
+      @inputArgs = {}
+
+      parseOneLevel = (arr, path)=>
+
+        returnObj = {}
+        if arr[0] is 'OPEN_BLOCK'
+          fragments = arr.slice(1,arr.length-1)
```
malamili: this is equivalent to ```arr.slice(1,-1)```

src/WeaverModel.coffee
```diff
@@ -0,0 +1,207 @@
...
+        else
+          fragments = arr.slice(2,arr.length-1)
+          @inputArgs['rootId'] = arr[0]
+
+
+        openedBlocks = 0
+
+        for fragment,i in fragments
+
+          if fragment == 'CLOSE_BLOCK'
```
malamili: following code conventions in coffeescript, try using keywords like is or isnt instead of ==

src/WeaverModel.coffee
```diff
@@ -0,0 +1,207 @@
...
+
+      pointer = @instance
+      pointer = pointer[p] for p in path.slice(0, -1)
+
+      pointer[path.slice(-1)[0]] = [] if pointer[path.slice(-1)[0]][0].indexOf('$') is 0
+
+      pointer[path.slice(-1)[0]].push(value)
+
+    @save = ->
+
```
malamili: There are so many empty lines, what is the rationale behind putting an empty line between each statement?

src/WeaverModel.coffee
```diff
@@ -0,0 +1,207 @@
...
+      @definition = definition
+
+class ModelInstance
+
+  constructor: (@modelDefinition, @inputArgs)->
+    @instance = {}
+    @instance[i] = j for i, j of @modelDefinition when i isnt 'inputArgs'
+
+    typeIsArray = Array.isArray || ( value ) -> return {}.toString.call( value ) is '[object Array]'
```
malamili: these functions are available in util.coffee

malamili:
I have to say its a bit complicated to follow what is happening, I'll checkout the branch and try to understand it better..

malamili:
More comments:

1. Models itself that are defined should be persisted, so let the class extend WeaveNode (Like for instance Weaver.Project)

2. The model.instance() method should return a WeaverNode in my opinion. There is so much duplication now reinvented in ModelInstance that has already been developed in WeaverNode, like the set, add and save functions.

3. There are so many if statements in parseOneLevel function, the complexity of this function is quite big  (https://en.wikipedia.org/wiki/Cyclomatic_complexity). Try to break it down in smaller functions. 

4. A bit more use of comments is recommended, for instance, what is inputArgs? Its defined someplace, and then used way down, and kind of still unclear what is does

src/WeaverModel.coffee
mingushead: `instance: ->`

I think that's bad naming on my part, _WeaverModel_ is supposed to represent the blueprint for a model (with variables), whereas _ModelInstance_ is supposed to represent an actual instance of that blueprint (with data where the blueprint variables were). That way, you define a Factory class once, and have as many specific factory instances as you want. The lexer and parser are required by the WeaverModel class, but not the ModelInstance class, hence they are defined outside of `instance: ->`

src/WeaverModel.coffee
```diff
@@ -0,0 +1,83 @@
...
+
+        else # do a recursive 'get' through child models
+          path = splitPath.slice(1).join('.')
+          arr =  (obj.get(path) for pred,obj of @relations[@definition[key].substr(1)].nodes) if @definition[key].charAt(0) is '@'
+          return util.flatten(arr, isFlattened)if isFlattened
+          arr
+
+      setProp: (key, val)->
+
+        return Error Weaver.Error.FILE_NOT_EXISTS_ERROR if not @definition[key]
```
gvdent: if @definition[key] is defined and false, this will also give an error. May never happen here, but it is good to be aware of. To check for non-existence do: something if not @definition[key]?

gvdent: Also FILE_NOT_EXISTS_ERROR does not seem appropriate here as this does not deal with files.

src/WeaverModel.coffee
```diff
@@ -0,0 +1,83 @@
...
+      constructor: (@nodeId)->
+        @definition = _def
+        staticProps = _statics
+        super(@nodeId)
+        @relation(key).add(rel) for rel in val for key,val of staticProps.rels
+
+        @setProp(key,val) for key,val of staticProps.attrs
+
+      get: (path, isFlattened = true)->
+        return @get(@definition[path].join('.')) if util.isArray(@definition[path])
```
gvdent: This is fine (upcoming comment about returns ;))

src/WeaverModel.coffee
```diff
@@ -0,0 +1,83 @@
...
+
+        # isFlattened:  default response should be flat array,
+        #               mark this false if property paths are required to be included in response
+        splitPath = path.split('.')
+        key = splitPath[0]
+
+        if splitPath.length is 1
+          # if @definition[key] is an array, they're relations, otherwise they're attributes
+          return (obj for pred,obj of @relations[@definition[key].substr(1)].nodes) if @definition[key].charAt(0) is '@'
+          return @attributes[@definition[key]]
```
gvdent: These lines would be a lot clearer with the if more explicit:
```
if @definition[key].charAt(0) is '@'
  (obj for pred,obj of @relations[@definition[key].substr(1)].nodes)
else
  @attributes[@definition[key]]
```

This way the control flow is more clear for the axe wielding maniac maintaining your code ;)

src/WeaverModel.coffee
```diff
@@ -0,0 +1,83 @@
...
+
+        else # do a recursive 'get' through child models
+          path = splitPath.slice(1).join('.')
+          arr =  (obj.get(path) for pred,obj of @relations[@definition[key].substr(1)].nodes) if @definition[key].charAt(0) is '@'
+          return util.flatten(arr, isFlattened)if isFlattened
+          arr
```
gvdent: Same comment as above, make the if more explicit.

In general: omit the return statement imo. Coffeescript returns the value of the last expression evaluated in a function anyway. Except when jumping out of a function earlier because of a stop condition or some error, explicitly adding it makes no difference (except for consistency when we'd start doing it everywhere).

src/util.coffee
```diff
@@ -1,6 +1,12 @@
 typeShouldBe = (type) -> (val) ->
   Object.prototype.toString.call(val) is type
 
+flatten = (arr) ->
```
gvdent: This would be an excellent candidate for a unit test to help clueless reviewers see what the purpose of this function is ;)

PR #40, opened 2017-02-16
WeaverModel creation tests + persistence refactor

mingushead:
- updates Gruntfile to include WeaverModel
- adds tests
- makes models persist to db properly

### DIFF
--- a/src/WeaverModel.coffee
+++ b/src/WeaverModel.coffee
@@ -6,50 +6,70 @@ circJSON    = require('circular-json')
 
 class WeaverModel extends Weaver.Node
 
-  constructor: (@name, @nodeId)->
+  constructor: (@nodeId)->
     super(@nodeId)
-    @set('name', @name) if @name
-
-  structure: (@definition)->
-
     @staticProps = {rels:{},attrs:{}}
+    @subModels = {}
+    @definition = {}
+
+  structure: (structure)->
+
+    for key,val of structure
+      if util.isArray(val)
+        @subModels[key] = val[1]
+        @definition[key] = val[0]
+      else @definition[key] = val
 
     # this attribute is used only for db storage purposes
     # - it should not be accessed directly.
     @set('definition', circJSON.stringify(@definition))
+    @set('subModels', circJSON.stringify(@subModels))
     @
 
   setStatic: (key, val)->
 
     throw new Error(Weaver.Error.CANNOT_SET_DEEP_STATIC) if util.isArray(@definition[key])
 
-    if @definition[key].charAt(0) is '@' # util.isArray(@definition[key])# add static relation for all model instances
-
+    if @definition[key].charAt(0) is '@'# add static relation for all model instances
       key = @definition[key].substr(1)
       @staticProps.rels[key] = @staticProps.rels[key] or []
       @staticProps.rels[key].push(val)
 
     else # add attribute static attribute for all model instances
       @staticProps.attrs[key] = val
+
+    @set('staticProps', circJSON.stringify(@staticProps))
     @
+
+  _loadFromQuery: (object)->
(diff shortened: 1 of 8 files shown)


src/WeaverModel.coffee
```diff
@@ -6,78 +6,126 @@ circJSON    = require('circular-json')
...
+                if splitPath.length is 1
+                  Promise.resolve(res)
+                else # path requires recursive get calls
+                  path = splitPath.slice(1).join('.') # to be used in next recursion
+                  proms = (obj.get(path, isFlattened) for obj in res)
+                  Promise.all(proms).then((arr)->
+                    if isFlattened
+                      Promise.resolve(util.flatten(arr, isFlattened))
+                    else
+                      Promise.resolve(arr)
```
gvdent: `Promise.all(proms).then((arr) ->` Will already ensure a promise is returned, so you can omit the `Promise.resolve`. 

This would then become:
```Promise.all(proms).then((arr)->
  if isFlattened
    util.flatten(arr)
  else
    arr
```
(Removed the second argument from util.flatten since that's not used)

src/WeaverModel.coffee
```diff
@@ -6,78 +6,126 @@ circJSON    = require('circular-json')
...
+
+  @loadModel: (id)-> # utility function to load a fully initialised Model from an id
+    Weaver.Node.load(id).then((node)=>
+      model = new Weaver.Model(id)
+      model._loadFromQuery(node)
+      Promise.resolve(model)
+    )
+
+  loadMember: (id)-> # utility function to load a fully initialised ModelMember from an id, given an associated Model
+    new Promise( (resolve, reject)=>
```
gvdent: In general, Promises should not be contructed explicitly (see http://bluebirdjs.com/docs/anti-patterns.html#the-explicit-construction-anti-pattern). In this code it leads to problems (specifically when an error is thrown during loading) because the Promise can forever be left hanging. 

Instead using existing Promise structure so any issues are also propagated up the chain to the point where they can be handled.

This code could be (depending on intentions):
```
loadMember: (id)->
  MemberClass = @buildClass()
  Weaver.Node.load(id).then((res)->
    member = new MemberClass(res.nodeId)
    member._loadFromQuery(res)
    member
  ).catch((err)-> #Note that this swallows the error. Possibly remove the entire catch and let the calling code handle rejections
    console.log(err)
  )
```

src/WeaverModel.coffee
```diff
@@ -6,78 +6,126 @@ circJSON    = require('circular-json')
...
+    @definition  = JSON.parse(@attributes.definition)
+    @staticProps = JSON.parse(@attributes.staticProps)
+    @subModels = JSON.parse(@attributes.subModels)
+    @structure(@definition)
+
+  @loadModel: (id)-> # utility function to load a fully initialised Model from an id
+    Weaver.Node.load(id).then((node)=>
+      model = new Weaver.Model(id)
+      model._loadFromQuery(node)
+      Promise.resolve(model)
```
gvdent: You're already in 'promise-land', so you can just return model here.

These are the general patterns:
```
  syncMethodReturningPromise: ->
    Promise.resolve(1)

  asyncMethodReturningPromise: ->
    syncMethodReturningPromise().then((res) ->
      res + 1
    )
```

The nice thing is that this method also ensures that if any promise in the chain is rejected, that rejection also prevents any futures which are dependant on that result from being evaluated, and the rejection will bubble up to the first rejection handler (`.catch((err) ->`) dealing with it.

mingushead: really helpful, thanks

src/WeaverModel.coffee
mingushead: +1

PR #44, opened 2017-03-13
Proposed solution

bastbijl:


src/WeaverRelation.coffee
```diff
@@ -14,12 +15,14 @@ class WeaverRelation
     Promise.resolve([])
 
   add: (node) ->
-    @nodes[node.id()] = node
-    @pendingWrites.push(Operation.Node(@parent).createRelation(@key, node.id()))
+    nodeId = node.id() if node.id
```
malamili: There is a better way in coffeescript
`nodeId = if node.id then node.id() else node`

src/WeaverRelation.coffee
```diff
@@ -14,12 +15,14 @@ class WeaverRelation
...
+    nodeId = node if not node.id
+    @nodes[nodeId] = node if node.id
```
malamili: This is going to break, apparently this function NEEDS a node always, right?

src/WeaverRelation.coffee
```diff
@@ -14,12 +15,14 @@ class WeaverRelation
...
+    @pendingWrites.push(Operation.Node(@parent).createRelation(@key, nodeId))
 
   remove: (node) ->
-    delete @nodes[node.id()]
-    @pendingWrites.push(Operation.Node(@parent).removeRelation(@key, node.id()))
+    delete @nodes[node.id() or node]
```
malamili: I think this throws an error if node is a string

test/WeaverNode.test.coffee
```diff
@@ -89,6 +89,17 @@ describe 'WeaverNode test', ->
       assert.isDefined(loadedNode.relation('comesBefore').nodes[bar.id()])
     )
 
+  it 'should add a new relation with static reference', ->
+    foo = new Weaver.Node()
+    bar = new Weaver.Node()
+    foo.relation('comesBefore').add(bar.id())
```
malamili: its not testing .add(node) with node a string
its not testing removing a node when node is a string

both will fail i think

malamili:
In general, I think these features are not worth the extra effort, possible extra fail paths, this overhead time of reviewing, documenting, testing etc..

I suggest next time to discuss these features on the JIRA board first before implementing, how small they may seem.

test/WeaverNode.test.coffee
bastbijl: it is testing the add with a string
remove test i will add

malamili: oh right, i see

PR #45, opened 2017-03-14
Feature/dnb 52 collect writes

bastbijl:
@malamili


src/WeaverNode.coffee
```diff
@@ -76,31 +76,39 @@ class WeaverNode
 
 
   # Go through each relation and recursively add all pendingWrites per relation AND that of the objects
-  _collectPendingWrites: (collected) ->
+  _collectPendingWrites: (pending, collected) ->
```
malamili: While i don't have much problems with having a batchSave functionality, i think this function got more complex (and it already was). Please move all functionality for the batchSave only in the batchSave function..so just loop over the nodes, concat all pendingWrites and be done it with.

bastbijl: agree

PR #49, opened 2017-03-15
Auth2

malamili:
I would say a pre-publish super alpha functionality
- WeaverUser is far from done in terms of functionality, this is a start
- Still some tests need to be made


test/WeaverUser.test.coffee
```diff
@@ -1,157 +1,170 @@
...
+      node.save()
+    ).catch((error) ->
+      # TODO: Assert error code
+      done()
     )
+    return
+
+
+  it 'should sign in the session if token is saved', ->
+    # TODO: Perhaps localforage is better for this instead of loki
```
carvilsi: Why? Maybe for async feature of localforage vs sync of lokijs?

malamili: dont know, i thought it might be overkill..but its good to go with loki i guess

src/WeaverACL.coffee
```diff
@@ -1,56 +1,114 @@
...
-
-  ## PUBLIC ##
+# For projects, users and nodes, you can specify which users and roles are allowed to read, and which
+# users and roles are allowed to modify.
+#
+# To support this type of security, each of these objects have an access control list,
+# implemented by this WeaverACL class.
+class WeaverACL
+
+  constructor: (user) ->
```
gvdent: I'd remove the user here and just add it to the ACL instance using the setUserWriteAccess method. This also gets rid of the issue where the constructor appears to sets this write access in the @_userWrite non-Map which is elsewhere used as an array. KISS

src/WeaverACL.coffee
```diff
@@ -1,56 +1,114 @@
...
+    @_publicWrite = false
+
+    # Locally these are objects, whereas the server expects arrays
+    # Converting to arrays before saving in save function
+    @_userReadMap  = {}
+    @_userWriteMap = {}
+    @_roleReadMap  = {}
+    @_roleWriteMap = {}
+
+    @_userWrite[user.id()] = true if user?
```
gvdent: @_userWriteMap?

src/WeaverACL.coffee
```diff
@@ -1,56 +1,114 @@
...
+  @load: (aclId) ->
+    coreManager = Weaver.getCoreManager()
+    coreManager.readACL(aclId)
+
+  save: ->
+    # Convert to array for all values that are true
+    trueKeys = (object) ->
+      (key for key, value of object when value)
+
+    @_userRead  = trueKeys(@_userReadMap)
```
gvdent: I'd create a data transfer object here as opposed to setting the _userRead, Write... etc members on this object. After a save this instance seems to be 'polluted' with what appear to be local variables for this function now.

src/WeaverProject.coffee
```diff
@@ -50,20 +46,21 @@ class WeaverProject extends Weaver.SystemNode
   getAllRelations:->
     Weaver.getCoreManager().getAllRelations(@id())
 
+  destroy: ->
+    coreManager = Weaver.getCoreManager()
```
gvdent: This line is duplicated something like 4 times. Can't coreManager be a (lazy-loaded) instance variable?

src/WeaverRole.coffee
```diff
@@ -1,21 +1,63 @@
...
+    (userId for userId, val of @_usersMap)
+
+  getRoles: ->
+    (roleId for roleId, val of @_rolesMap)
+
+  delete: ->
+    coreManager = Weaver.getCoreManager()
+    coreManager.deleteRole(@id()).then(=>
+      @_deleted = true
+      return
```
gvdent: What does including the explicit return here add functionally?

test/Integration.test.coffee
```diff
@@ -0,0 +1,170 @@
...
+
+      # Signup with a different user and assess that that user has no right to read this project
+      john = new Weaver.User('john', "secretSauce", "john@doe.com")
+      john.signUp()
+      reg.john = john # Save for later
+
+    ).then(->
+
+      # Now we should not be able to load the private node
+      Weaver.Node.load('adminPrivateNode')
```
gvdent: There is no assertion here that the error is actually thrown, if no rejection occurs the test will pass as well.

test/Integration.test.coffee
```diff
@@ -0,0 +1,170 @@
...
+      # Should now be able to read this node
+      Weaver.Node.load('adminPrivateNode')
+
+    ).then((node) ->
+
+      assert.equal(node.get('name'), 'Foo')
+
+      # Assert that writing is still not allowed
+      node.set('name', 'Sonic')
+      node.save()
```
gvdent: Check that an error is given here (it is not asserted that the catch clause below has been entered).

test/Integration.test.coffee
```diff
@@ -0,0 +1,170 @@
...
+
+      # Change the name
+      node.set('name', 'Sonic')
+      node.save()
+
+    ).then((node) ->
+
+      assert.equal(node.get('name'), 'Sonic')
+
+    ).then(->
```
gvdent: And then and then and then... sounds like my 5 year old ;)

test/WeaverNode.test.coffee
```diff
@@ -120,8 +120,14 @@ describe 'WeaverNode test', ->
     )
 
   it 'should give an error if node already exists', ->
-    node = new Weaver.Node('a')
-    node.save().then((res) ->
+    return # TODO implement this
```
gvdent: Instead of using return before failing checks, use 
```
it.skip 'should give an error...'
```
That way a disabled test shows up in the npm test run overview as pending instead of this showing a passing test.

test/WeaverProject.test.coffee
```diff
@@ -16,27 +15,31 @@ describe 'WeaverProject Test', ->
...
       expect(p.id()).to.equal(project.id())
-      project.destroy()
       done()
     )
     return
 
 
   it 'should create projects with attributes', (done) ->
+    done()
+    return # TODO: Implement this
```
gvdent: skip

test/WeaverProject.test.coffee
```diff
@@ -16,27 +15,31 @@ describe 'WeaverProject Test', ->
...
       expect(loadedProject.get("name")).to.equal("test")
-      project.destroy()
       done()
     )
     return
 
 
   it 'should delete projects', (done) ->
+    done()
+    return # TODO: Project loading does not work, implement this
```
gvdent: skip

test/WeaverProject.test.coffee
```diff
@@ -45,34 +48,29 @@ describe 'WeaverProject Test', ->
     ).then(->
       Weaver.Project.load(test.id())
     ).catch((error) ->
+      console.log(error)
```
gvdent: Please don't write to console.log on the expected execution path.

test/WeaverUser.test.coffee
```diff
@@ -1,157 +1,170 @@
...
-    weaverUser = new Weaver.User()
-    weaverUser.logOut()
-    .then( ->
+      # Assert email and username are set, while password is not set
+      assert.equal(loadedUser.username, username)
+      assert.equal(loadedUser.email, "centaurus@univer.se")
+      assert.isTrue(not loadedUser.password?)
+      done()
     )
+    return
```
gvdent: Why is this return here? Is it a requirement or some style decree I am not aware of?

test/WeaverUser.test.coffee
```diff
@@ -1,157 +1,170 @@
...
-      assert.equal(err.code, Weaver.Error.USERNAME_NOT_FOUND)
-    )
 
-  it 'should fails trying logOut action specifying non loggedin user', ->
-    weaverUser = new Weaver.User()
-    weaverUser.logOut('andromeda')
-    .then().catch((err) ->
-      assert.equal(err.code, Weaver.Error.USERNAME_NOT_FOUND)
-    )
+  it 'should create the admin user upon initialization', (done) ->
```
gvdent: 107,150:s/it/it.skip/

Man I'm really going to look like an ass if the .skip is not actually the right suffix.

src/WeaverACL.coffee
gvdent: After checking the server code, this may not be a valid comment.

src/WeaverACL.coffee
malamili: Oeps, thats actually a bug, it should have been the map in the constructor...nevertheless, good point, changing it

src/WeaverACL.coffee
malamili: haha yes, bug! Oh well, this class hasnt been tested at all..next TODO

src/WeaverACL.coffee
malamili: Yep, valid comment. I thought this was ugly..transfer object sounds like a good Idea...i want to do it for WeaverRole and WeaverProject too

src/WeaverProject.coffee
malamili: I made it a scope variable then..

src/WeaverRole.coffee
malamili: removed

test/Integration.test.coffee
malamili: good catch!

test/Integration.test.coffee
malamili: 👍

test/Integration.test.coffee
malamili: haha, fixed!

test/WeaverNode.test.coffee
malamili: nice!

test/WeaverUser.test.coffee
malamili: Goood question, this has been bugging me. If i remove the return, this happens:

     Error: Resolution method is overspecified. Specify a callback *or* return a Promise; not both.

If i then remove the done() (because it is returning a promise)..the test never finishes. Can you figure out why? SO now i always use done() and return

test/WeaverProject.test.coffee
malamili: 👍

test/WeaverProject.test.coffee
malamili: 👍

test/WeaverProject.test.coffee
malamili: 👍

test/WeaverUser.test.coffee
malamili: haha

test/WeaverUser.test.coffee
gvdent: Ah right... you can return the promise and not use the done argument, but that requires rewriting the test as you don't get the implicit 'assert that done() is reached' that the parameterized test has. 

Leave it be :)

PR #54, opened 2017-03-29
Add setScheduler function

Gielert:
The setScheduler function is needed for the weaver-sdk-angular. This would allow the angular-sdk to set a callback function on the Promise library so we can trigger the digest cycle in Angular.
See code for changes.


malamili:
Congratulations, this must be the first PR with just 2 lines 👍

PR #56, opened 2017-03-30
Is loaded check

bastbijl:
@malamili


src/WeaverNode.coffee
```diff
@@ -8,6 +8,8 @@ class WeaverNode
   constructor: (@nodeId) ->
     # Generate random id if not given
     @nodeId = cuid() if not @nodeId?
+    @_stored = false
+    @_loaded = true
```
malamili: I still believe this should be false

src/WeaverNode.coffee
```diff
@@ -26,22 +28,27 @@ class WeaverNode
   _loadFromQuery: (object, Constructor) ->
     Constructor = Constructor or WeaverNode
     @nodeId     = object.nodeId
+    @_stored    = true
```
malamili: much better to move these to Weaver.load function, that way, you dont have to hacky set it back to false on line 39

src/WeaverNode.coffee
```diff
@@ -26,22 +28,27 @@ class WeaverNode
...
-    @._clearPendingWrites()
+    @._clearPendingWrites(true)
     @
 
 
   # Node creating for in queries
   @get: (nodeId, Constructor) ->
     Constructor = WeaverNode if not Constructor?
     node = new Constructor(nodeId)
+    node._stored = true
```
malamili: This is not true, the Node might not be stored.

malamili: Also, I believe its better to remove them here too. There are only three places where you want these to change:
- constructor -> default to not loaded and not stored (although in reality you dont know)
- in the save, only stored to true
- in the load, both stored and loaded to true

src/WeaverNode.coffee
```diff
@@ -118,12 +125,13 @@ class WeaverNode
 
 
   # Clear all pendingWrites, used for instance after saving or when loading a node
-  _clearPendingWrites: ->
+  _clearPendingWrites: (stored)->
```
malamili: please don't put this here, this function is complex, adding more stuff makes it difficult to understand. If you MUST  traverse nodes, do it again in a separate function. Better to avoid, do we really need this or is it a result of bad design?

src/WeaverNode.coffee
bastbijl: what about the loadFromQuery?

src/WeaverNode.coffee
bastbijl: nodes must be traversed, because you dont know upfront if the save will succeed, it is not the result of bad design

it is ugly though that the clearPendingWrites is called from the get()

src/WeaverNode.coffee
bastbijl: never mind this last comment, i know what about

src/WeaverNode.coffee
malamili: hmm, maybe. But i what i meant, instead of affecting the behaviour of clearPendingWrites with arguments, for complexity sake, better to have this function:
- _allNodesStored: -> recursive set stored, and use this function after succesful save on line 148

PR #57, opened 2017-04-06
Json webtoken

Gielert:
Authenticating using a JSON Webtoken is now supported in the SDK. Weaver.signIn(username, password) still works, but you will also be able to supply a token as the argument for signIn (Weaver.signIn(token)). If the token is still valid, the user will be signed in.

See code for changes

This pull request is in line with https://github.com/weaverplatform/weaver-server/pull/63


src/Weaver.coffee
```diff
@@ -64,8 +64,15 @@ class Weaver
   signOut: ->
     @coreManager.signOutCurrentUser()
 
-  signIn: (username, password) ->
-    @coreManager.signInUser(username, password)
+  # args may be username/password or a token
+  signIn: (args...) ->
```
malamili: See my comments on weaver-server PR about this being bad practice, its better to split this up in two functions.

Gielert: What if we do make signing in like this

```js
signIn: ->
  withUsername: @coreManager.signInUsername.bind(@coreManager)
  withToken: @coreManager.signInToken.bind(@coreManager)
```
With the above example we can login as the example below:
```js
Weaver.signIn().withUsername("username", "password")
Weaver.signIn().withToken("token")
```

malamili: There is no need, you usually do this when you pipe data around. Its cleaner to just do this:

```
Weaver.signInWithUsername("username", "password")
Weaver.signInWithToken("token")
```

Gielert:
As for storing the tokens in the LocalStorage. If you are in a web environment you can just use the localStorage object to store the token, but when we're in a node environment we need something like localForage (I don't know if we can use this for node though).
But for now I think this will do the trick.

test/WeaverUser.test.coffee
```diff
@@ -34,6 +34,22 @@ describe 'WeaverUser Test', ->
     )
     return
 
+  it 'should sign in using a token', ->
```
malamili: Looks good! Request: Can you test if failed signIn using a token?

PR #74, opened 2017-05-08
Remove WeaverRoot

Gielert:
WeaverRoot is now gone, and replaced by the static getInstance or getCoreManager functions in Weaver.
Require of Weaver objects in done after the exports of Weaver to solve the circular dependency issue. This is just a slightly cleaner way to make Weaver available in every Weaver object.
Weaver objects are also available from the weaver instance.
So like this:
```javascript
const weaver = new Weaver()
const node = new weaver.Node()
// But objects are also still available like this
const staticNode = new Weaver.Node()
```

**NOTE!!!**
All getClass functions in Weaver objects are gone. Since they are now available through the static reference and the instance reference.


src/Weaver.coffee
```diff
@@ -1,37 +1,33 @@
...
-
-  getClass: ->
-    Weaver
-  @getClass: ->
-    Weaver
+class Weaver
 
   constructor: ->
 
     if Weaver.weaver?
```
gvdent: Should this now not be `if Weaver.instance?`

Gielert: You are right. Fixed it :)

malamili:
Awesome!!!

PR #79, opened 2017-05-16
New flag in weaver query and a test

bastbijl:
@gvdent @malamili

### DIFF
--- a/src/WriteOperation.coffee
+++ b/src/WriteOperation.coffee
@@ -13,7 +13,7 @@ Action.CREATE_ATTRIBUTE       = define 'create-attribute',    ['timestamp', 'id'
 Action.UPDATE_ATTRIBUTE       = define 'update-attribute',    ['timestamp', 'id', 'key', 'value', 'datatype']
 Action.REMOVE_ATTRIBUTE       = define 'remove-attribute',    ['timestamp', 'id', 'key']
 Action.CREATE_RELATION        = define 'create-relation',     ['timestamp', 'from', 'key', 'to', 'id' ]          # only one relation with this key can exist between these two node ids
-Action.UPDATE_RELATION        = define 'update-relation',     ['timestamp', 'from', 'key', 'oldTo', 'newTo' ]
+Action.UPDATE_RELATION        = define 'update-relation',     ['timestamp', 'from', 'key', 'oldTo', 'newTo', 'id' ]
 Action.REMOVE_RELATION        = define 'remove-relation',     ['timestamp', 'from', 'key', 'to' ]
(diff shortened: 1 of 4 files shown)


src/WriteOperation.coffee
```diff
@@ -13,7 +13,7 @@ Action.CREATE_ATTRIBUTE       = define 'create-attribute',    ['timestamp', 'id'
 Action.UPDATE_ATTRIBUTE       = define 'update-attribute',    ['timestamp', 'id', 'key', 'value', 'datatype']
 Action.REMOVE_ATTRIBUTE       = define 'remove-attribute',    ['timestamp', 'id', 'key']
 Action.CREATE_RELATION        = define 'create-relation',     ['timestamp', 'from', 'key', 'to', 'id' ]          # only one relation with this key can exist between these two node ids
-Action.UPDATE_RELATION        = define 'update-relation',     ['timestamp', 'from', 'key', 'oldTo', 'newTo' ]
+Action.UPDATE_RELATION        = define 'update-relation',     ['timestamp', 'from', 'key', 'oldTo', 'newTo', 'id' ]
```
malamili: Whats this?

test/WeaverQuery.test.coffee
```diff
@@ -101,6 +101,31 @@ describe 'WeaverQuery Test', ->
       )
     )
 
+  it 'should not return relations', ->
```
malamili: noice!!

src/WriteOperation.coffee
gvdent: @bastbijl What is the impact of this change exactly? I don't see any related changes except those in the weaver-database-connector. Can older SDK versions still use the update relation without an id?

test/WeaverQuery.test.coffee
```diff
@@ -101,6 +101,31 @@ describe 'WeaverQuery Test', ->
...
+      )
+    ).then(->
+      new Weaver.Query()
+      .noRelations()
+      .find().then((nodes) ->
+        expect(nodes.length).to.equal(2)
+        checkNodeInResult(nodes, 'a')
+        checkNodeInResult(nodes, 'b')
+      )
+    )
```
gvdent: Minor nitpick: ideally these would be different testcases, so that when one fails it is immediately obvious where the error is.

src/WriteOperation.coffee
bastbijl: the impact of this specific change is null, this json signature doesn't do much

but the bigger picture: the impact of adding this requirement is huge, an update of the sdk is needed

bastbijl: So no, the protocol version should be increased.

gvdent: Ok, so old clients will not be able to use weaver-server with database-connector 0.1.5?

PR #80, opened 2017-05-16
Remove unnecessary dependencies

Gielert:
Removed two unnecessary dependencies.
Also we don't expect the dev to have gulp and coffeescript globally installed now.
Updating dependencies to latest stable version can be done with:
```
$ npm run update-deps
```
or
```
$ yarn update-deps
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,8 +19,6 @@
     "circular-json": "^0.3.1",
     "config": "^1.24.0",
     "cuid": "~1.3.8",
-    "fs-readfile-promise": "^3.0.0",
-    "fs-writefile-promise": "^2.0.0",
     "localforage": "^1.5.0",
     "lodash": "^4.17.4",
     "lokijs": "^1.4.2",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -53,7 +52,8 @@
     "node": ">=6.9"
   },
   "scripts": {
-    "prepublish": "coffee -o lib -c src && gulp dev",
+    "prepublish": "./node_modules/coffee-script/bin/coffee -o lib -c src && ./node_modules/gulp/bin/gulp.js dev",
+    "update-deps": "./node_modules/david/bin/david.js u -i gulp",
```
gvdent: Nice :)

One thing though: `npm run` and `yarn run` add `./node_modules/.bin` to the path, so coffee wasn't required globally already. These lines can read 
```
"prepublish": "coffee -o lib -c src && gulp dev",
"update-deps": "david u -i gulp",
```

Unless that's an artifact of my setup.

PR #83, opened 2017-05-19
Generate snapshot write operations

bastbijl:
@gvdent @malamili


carvilsi:
@bastbijl what's the difference between snapshotProject and dumpHistory?

gvdent:
@carvilsi technically: writeHistory dumps the trackerdb info, snapshot derives from the virtuoso database. 

conceptually: writeHistory contains all history used to create the current database state, snapshot only stores the current state without any history.

bastbijl:
@carvilsi dump history reads from the trackerdb and reproduces the whole history, snapshot creates write operations based on the content of the graph database. The minimal set of write operations to reproduce the snapshot current graph database state. So no history.

carvilsi:
Thanks @bastbijl and @gvdent

PR #86, opened 2017-05-24
fixing issue related with https with self signed certs

carvilsi:


src/SocketController.coffee
```diff
@@ -11,6 +11,7 @@ class SocketController
 
     @options = @options or defaultOptions
     @options.query = "sdkVersion=#{pjson.version}&requiredServerVersion=#{pjson.com_weaverplatform.requiredServerVersion}"
+    @options.rejectUnauthorized = false
```
gvdent: This option should not be set to false in the default SDK, since it allows for MITM attacks. Maybe a separate release for those who like to live on the edge (or self sign certificates).

carvilsi: I propose to add a config var to set this value to true/false.

Gielert: Yes, this would be a better option

carvilsi:
Added at config: `rejectUnauthorized` and set to *true* by default.
Added `WEAVER_REJECT_UNAUTHORIZED` as environment variable
Small require cleaning on File test.

dist/weaver-sdk.full.js
```diff
@@ -99666,6 +99666,7 @@ module.exports={
       };
       this.options = this.options || defaultOptions;
       this.options.query = "sdkVersion=" + pjson.version + "&requiredServerVersion=" + pjson.com_weaverplatform.requiredServerVersion;
+      this.options.rejectUnauthorized = false;
```
gvdent: I think you need to run npm-prepublish one more time, since this differs from the coffee code.

src/SocketController.coffee
```diff
@@ -11,6 +12,7 @@ class SocketController
 
     @options = @options or defaultOptions
     @options.query = "sdkVersion=#{pjson.version}&requiredServerVersion=#{pjson.com_weaverplatform.requiredServerVersion}"
+    @options.rejectUnauthorized = config.get('weaver.rejectUnauthorized')
```
gvdent: 👍

PR #89, opened 2017-05-29
Https self signed cert

carvilsi:
removing `rejectUnauthorized` as config param and passing this value as option on weaver.connect, the default value is set to true:
`weaver.connect(config.url,{rejectUnauthorized: false}).then()...`

The `rejectUnauthorized` is maintained at config file for testing purposes.

### DIFF
--- a/src/SocketController.coffee
+++ b/src/SocketController.coffee
@@ -2,17 +2,16 @@
 io       = require('socket.io-client')
 Promise  = require('bluebird')
 pjson    = require('../package.json')
-config   = require('config')
 
 class SocketController
 
   constructor: (@address, @options) ->
     defaultOptions =
       reconnection: true
+      rejectUnauthorized: true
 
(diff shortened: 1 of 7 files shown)


src/SocketController.coffee
```diff
@@ -2,17 +2,16 @@
...
-config   = require('config')
 
 class SocketController
 
   constructor: (@address, @options) ->
     defaultOptions =
       reconnection: true
+      rejectUnauthorized: true
 
     @options = @options or defaultOptions
```
gvdent: I think it's appropriate to set `@options.reconnection = true`. That connection option is not intended to be modified by the caller afaik. Currently options are passed which override the defaults here, and thus reconnection is never set to true.

If it is intended to be an option, ensure that the value is actually set, and if not set it to true. But I think that unnecessary complexity, even simple as it is.

PR #90, opened 2017-06-08
adding test for NoSQL injection attempts

carvilsi:
testing if there is an attempt to sigIn using NoSQL injection


gvdent:
More tests more better 👍

PR #91, opened 2017-06-09
Issue 82 project list

gvdent:
Fixes Weaver.Project.list() to return functional Projects, which can be destroyed and used in useProject etc.

### DIFF
--- a/src/WeaverProject.coffee
+++ b/src/WeaverProject.coffee
@@ -5,10 +5,9 @@ class WeaverProject
 
   @READY_RETRY_TIMEOUT: 200
 
-  constructor: (@name, @projectId) ->
+  constructor: (@name, @projectId, @_stored = false) ->
     @name = @name or 'unnamed'
     @projectId = @projectId or cuid()
-    @_stored = false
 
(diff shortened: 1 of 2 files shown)


src/WeaverProject.coffee
```diff
@@ -5,10 +5,9 @@ class WeaverProject
 
   @READY_RETRY_TIMEOUT: 200
 
-  constructor: (@name, @projectId) ->
+  constructor: (@name, @projectId, @_stored = false) ->
```
malamili: nice!

src/WeaverProject.coffee
```diff
@@ -59,6 +58,8 @@ class WeaverProject
     Weaver.getCoreManager().getACL(@projectId)
 
   @list: ->
-    Weaver.getCoreManager().listProjects()
+    Weaver.getCoreManager().listProjects().then((list) ->
+      ( new Weaver.Project(p.id, p.name, true) for p in list )
```
malamili: I think we're missing some more fields to be set such as acl

PR #100, opened 2017-06-25
Change password and various WeaverUser additions

malamili:


src/CoreManager.coffee
```diff
@@ -125,24 +125,20 @@ class CoreManager
...
-    payload =
-      userId:   user.userId
-      username: user.username
-      password: user.password
-      email:    user.email
+    update = {}
+    update[key] = value for key, value of user
 
-    @POST("user.signUp", payload, "$SYSTEM")
+    @POST("user.signUp", update, "$SYSTEM")
```
gvdent: At this point, isn't update just equal to user?

PR #104, opened 2017-06-28
Clone default keycap

malamili:


src/WeaverNode.coffee
```diff
@@ -111,12 +111,14 @@ class WeaverNode
           Constructor.load(id).then((node)->
             node.clone({}, self).then((node)->
               clone.relation(key).add(node)
-              Promise.resolve(clone)
+              return Promise.resolve(clone)
```
gvdent: This function is broken: the result of the Constructor.load(id)... promise chain is not integrated into the main function. These can still be unresolved while the latter return Promise.resolve(clone) does fire, resulting in an incomplete clone.

I think.

PR #105, opened 2017-06-30
History

carvilsi:


carvilsi:
Adding 'from' to history query and allowing a node ID or an array of those as node input to getHistory

PR #110, opened 2017-07-20
Load all nodes in a relation

Gielert:
We can now load all nodes in a relation by doing

```js
node.relation('someRelation').load()
.then((nodes) => {
  //array of loaded nodes
})
```


malamili:
I like them clean short and simple, nice!

PR #112, opened 2017-08-08
Sequential node save

gvdent:


src/WeaverNode.coffee
```diff
@@ -226,12 +226,32 @@ class WeaverNode
...
-      @_clearPendingWrites()
-      @_setStored()
-      @
+    cm = Weaver.getCoreManager()
+
+    sp = cm.operationsQueue.then(=>
+      writes = @_collectPendingWrites()
+
+      cm.executeOperations(writes, project).then(=>
+        @_clearPendingWrites()
```
gvdent: Bug: If you perform some operations, call save (which experiences a delay for some reason), then perform some other operations, after which the original save resolves; the write operations are cleared even if no save on the new operations have been called yet.

gvdent: This bug was already previously present but may occur more often now that all saves are sequential

PR #122, opened 2017-08-18
No attr rel defaults

gvdent:
Makes noRelations and noAttributes default for query so the default uses the new template

### DIFF
--- a/src/WeaverQuery.coffee
+++ b/src/WeaverQuery.coffee
@@ -19,7 +19,6 @@ class WeaverQuery
     @_select       = []
     @_noRelations  = true
     @_noAttributes = true
-    @_liveIndividualsOnly = true
     @_count        = false
     @_hollow       = false
     @_limit        = 99999
--- a/src/WeaverQuery.coffee
+++ b/src/WeaverQuery.coffee
@@ -19,6 +19,7 @@ class WeaverQuery
     @_select       = []
     @_noRelations  = true
     @_noAttributes = true
+    @_liveIndividualsOnly = true
     @_count        = false
     @_hollow       = false
     @_limit        = 99999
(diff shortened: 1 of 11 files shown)


test/WeaverQuery.test.coffee
```diff
@@ -187,7 +188,7 @@ describe 'WeaverQuery Test', ->
 
     a.save().then(->
       new Weaver.Query()
-      .noRelations()
+      .noRelationNodes()
```
mingushead: This is unnecessary, right? Should I add one more duplicate test without this line to make sure the defaults stay set up correctly?

src/WeaverQuery.coffee
```diff
@@ -209,5 +222,10 @@ class WeaverQuery
   nativeQuery: (query)->
     Weaver.getCoreManager().nativeQuery(query, Weaver.getInstance().currentProject().id())
 
+  includeAttributesAndRelations: ->
```
malamili: what does this mean? How is it different from noRelations and noAttributes?

malamili:
Whats the purpose of ```includeAttributesAndRelations```?

src/WeaverQuery.coffee
gvdent: Ugh, this bothers me as well. The options differ right now in the following manner:

includeAttributesAndRelations allows loading attribute and relation nodes directly. If this option is skipped loading a node with an id that is not an individual will fail (this is new template optimization behaviour).

As far as I understand noRelation and noAttributes determine whether the attribute/relations "children" nodes of actual queried nodes are loaded. So, querying for the 'individual-1' node will also load the 'individual-1-relation-1', 'individual-1-attribute-1' and 'individual-1-attribute-2' nodes. Correct me if I'm wrong on the expected behaviour but that I how I understood it from the tests.

I'm still pondering what to do with the options: whether to merge them or rename them in a way that their intention is more obvious.

malamili: Oeps! Haha you understood it wrong..noRelation and noAttributes mean that these nodes dont get returned as root nodes (so not meaning the children of the individuals)..

malamili: I think noRelation and noAttributes is what `NOT includeAttributesAndRelations` means right now

PR #124, opened 2017-08-22
Weaver Relation Fix

Gielert:
Important fix for relations. All nodes that were loaded through the load function in WeaverRelation were not updates inside the relation itself. This fixes that issue.


malamili:
nice!

PR #125, opened 2017-08-22
Weav 127 download dump

aal89:


aal89:
@gvdent The implementation is now like you asked. The second parameter of the getSnapshot is optional which tells the server to output a zip instead of plain text. All other tests also pass (also including this new on, ofc hehe)

PR #126, opened 2017-08-22
List users for users

gvdent:
Adds a call to the server to retrieve users which also have access to the current project

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "email": "mohamad@sysunite.com"
   },
   "com_weaverplatform": {
-    "requiredServerVersion": "~3.0.3 || ^3.0.4-beta.0"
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
     "email": "mohamad@sysunite.com"
   },
   "com_weaverplatform": {
-    "requiredServerVersion": "~3.0.3 || ^3.0.4-beta.0"
+    "requiredServerVersion": "~3.0.5 || ^3.0.5-beta.0"
```
Gielert: We're currently on 3.0.5-beta.2 IIRC

Gielert: ah nvm. This is required server-version. Herpaderp....

PR #129, opened 2017-08-23
Weav 138 readonly

aal89:
http://jira.sysunite.com/browse/WEAV-138

### DIFF
--- a/test/WeaverProject.test.coffee
+++ b/test/WeaverProject.test.coffee
@@ -73,15 +73,12 @@ describe 'WeaverProject Test', ->
     p.freeze().then(->
       a = new Weaver.Node()
       a.save().then(->
-        # Writing a node after freeze should not be possible
-        assert(false)
+        assert(false, "Writing a node after freeze should not be possible")
       ).catch((err)->
-        # Writing this node should be impossible
-        assert(true)
+        assert.include(err.message, "Project is frozen")
       )
     ).catch((err)->
-      # Default case is to fail this test
-      assert(false)
+      assert(false, "Default case is to fail this test, project wasn't frozen? " + err.message)
     )
 
   it 'should unfreeze a project making writing possible', ->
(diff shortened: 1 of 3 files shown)


test/WeaverProject.test.coffee
```diff
@@ -68,6 +68,38 @@ describe 'WeaverProject Test', ->
       weaver.useProject(p)
     )
 
+  it 'should freeze a project making writing impossible', ->
```
gvdent: Great test coverage of the new functionality :)

test/WeaverProject.test.coffee
```diff
@@ -68,6 +68,38 @@ describe 'WeaverProject Test', ->
...
+      a.save().then(->
+        # Writing a node after freeze should not be possible
+        assert(false)
+      ).catch((err)->
+        # Writing this node should be impossible
+        assert(true)
+      )
+    ).catch((err)->
+      # Default case is to fail this test
+      assert(false)
```
gvdent: This statement is not necessary, if the last statement of a test is a promise and it rejects the test fails automatically.

test/WeaverProject.test.coffee
```diff
@@ -68,6 +68,38 @@ describe 'WeaverProject Test', ->
...
+    )
+
+  it 'should unfreeze a project making writing possible', ->
+    p = weaver.currentProject()
+    p.unfreeze().then(->
```
gvdent: This tests can be rewritten with more explicit feedback if something fails. We're using chai, so you have a couple of options (assert(boolean) will only fail with an assertion failed message, but you will not know which of the 3 asserts it was that failed).

I'd rewrite this test something along the lines of 
```
it 'should unfreeze a project making writing possible', ->
  p = weaver.currentProject()
  p.unfreeze().then(->
    (new Weaver.Node()).save()
  )
```
The messaging will pretty much remain except that the original error message in case of a rejection would be printed. Since in the passing case the promise is resolved as expected, the test would pass.

test/WeaverProject.test.coffee
```diff
@@ -68,6 +68,38 @@ describe 'WeaverProject Test', ->
       weaver.useProject(p)
     )
 
+  it 'should freeze a project making writing impossible', ->
+    p = weaver.currentProject()
+    p.freeze().then(->
```
gvdent: Since we use chai-as-promised, this test could be rewritten as:
```
it 'should freeze a project making writing impossible', ->
  weaver.currentProject().freeze().then(->
    (new Weaver.Node()).save().should.be.rejected
  )
```

PR #130, opened 2017-08-24
WEAV 163 Renaming projects

aal89:
http://jira.sysunite.com/browse/WEAV-163

### DIFF
--- a/test/WeaverProject.test.coffee
+++ b/test/WeaverProject.test.coffee
@@ -140,3 +140,12 @@ describe 'WeaverProject Test', ->
       expect(p.id()).to.equal('helloworld_dupe')
       p.destroy()
     )
+
+  it 'should rename a project on the server', ->
+    p = weaver.currentProject()
+    p.rename('rename_test')
+    .then(->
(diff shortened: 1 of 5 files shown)


test/WeaverProject.test.coffee
```diff
@@ -140,3 +140,12 @@ describe 'WeaverProject Test', ->
       expect(p.id()).to.equal('helloworld_dupe')
       p.destroy()
     )
+
+  it 'should rename a project on the server', ->
+    p = weaver.currentProject()
+    p.rename('rename_test')
+    .then(->
+      expect(p.name).to.equal('rename_test')
```
gvdent: This test does not cover the roundtrip, since the local instance may be the only thing modified. To check that the name on the server has also been modified, use Weaver.Project.list() and check the project returned with the same project id for the correct name.

aal89:
NOTE: the last commit, I merged the earlier WEAV-138 issue into this branch. So merge that one prior merging this branch in order to avoid conflicts. Those PR's have the same base, hence the merge.

PR #135, opened 2017-09-07
Weav 128 upload project

aal89:
http://jira.sysunite.com/browse/WEAV-128


aal89:
For the recent strange behaviour, please see: http://jira.sysunite.com/browse/WEAV-176. I documented my research into larger zip files and it's executing through the SDK -> server and finally the connector. I believe the changes are now definite.

PR #140, opened 2017-09-09
useConstructor function

malamili:


malamili:
Pair programming with @bastbijl

PR #144, opened 2017-09-20
Concurrent save

mingushead:
Allows ignoring of out-of-date attributes updates during `Weaver.Node.save()` operations, as a workaround until a better system of concurrent interaction is devised.

### DIFF
--- a/test/WeaverNode.test.coffee
+++ b/test/WeaverNode.test.coffee
@@ -136,7 +136,7 @@ describe 'WeaverNode test', ->
       assert.equal(loadedNode.get('time').toJSON(), date.toJSON())
     )
 
-  it 'should increment an exiting number attribute', ->
+  it 'should increment an existing number attribute', ->
     node = new Weaver.Node()
     node.set('length', 3)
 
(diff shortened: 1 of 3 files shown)


test/WeaverNode.test.coffee
```diff
@@ -355,3 +355,44 @@ describe 'WeaverNode test', ->
...
+    ).then((res)->
+      assert.equal(res.get('name'),'A_')
+    ).then(->
+      ay.set('name', 'Ay')
+      ay.save(null,{ignoresOutOfDate: true})
+      aay.set('name','Aay')
+      aay.save(null,{ignoresOutOfDate: true})
+    ).catch((err)->
+      console.log(err)
+      assert.equal('money','happiness')
```
gvdent: This last .catch clause to fail the test is not needed, it a test returns a promise (which this does), it will automatically fail if that promise is rejected.

test/WeaverNode.test.coffee
```diff
@@ -355,3 +355,44 @@ describe 'WeaverNode test', ->
     ).then((node) ->
       assert.isDefined(node.relation('to').nodes[cloned.id()])
     )
+
+  it 'should handle concurrent saves from multiple references, when the flag is passed', ->
```
gvdent: If this currently-under-discussion PR is going to eventually be merged, please also create a test that shows the update fails if the flag is not passed to illustrate the difference (if no such test exists yet)

malamili:
TBH the code is very hacky getting stuff from the error message (not to mention its a postgres error message, might fail very quickly)..i suggest we merge this PR but very quickly move to a connector impl

gvdent:
Closing this PR for now, the branch will live and the feature will be further developed before a new one is opened.

PR #147, opened 2017-10-02
Weav 191 fix clone

gvdent:
Clone functionality was fixed in 0.0.21-SNAPSHOT of the postgresql connector.


malamili:
Should develop first be merged into this? it has changes from another PR

PR #148, opened 2017-10-02
Added functionality to use .hasRelationOut with strings as input

mathieub93:
@gvdent


src/WeaverQuery.coffee
```diff
@@ -122,8 +122,12 @@ class WeaverQuery
...
       @_addCondition(key, '$relOut', node)
-    else
-      @_addCondition(key, '$relOut', if node.length > 0 then (i.id() or i for i in node) else ['*'])
+    else if node.length > 0
+      nodeIds = []
+      nodeIds.push i for i in node when typeof i is 'string'
+      nodeIds.push i.id() || i for i in node when typeof i isnt 'string'
+      @_addCondition(key, '$relOut', nodeIds)
+    else ['*']
     @
```
malamili: Looks like functionality is broken now. The else ['*'] should have been (i think:) 
@_addCondition(key, '$relOut',  ['*'])

Did you run all tests to see if this hasnt broken anything?

Also, use ( ) in the push function by our style convention (see https://github.com/weaverplatform/weaver-server/blob/develop/coding-style.md)

Its also better to explicitly check if input is a node instead of 'not typeof string', because you'll get unspecified undefined errors when supplying a double or boolean on the .id(). Check util.coffee and use it instead

Lastly, the traversing two times with the pushes is'nt very readable, how about this:

```
getId = (input) ->
   if isString(input)
      return input
   else if isNode(input)
      return input.id()
   else
     throw new Error("Unsupported type")   

 @_addCondition(key, '$relOut', (getId(input) for input in node))
```

src/WeaverQuery.coffee
```diff
@@ -109,21 +109,29 @@ class WeaverQuery
   greaterThanOrEqualTo: (key, value) ->
     @_addCondition(key, '$gte', value)
 
+  getId = (node) ->
```
malamili: Declare this function outside the WeaverQuery class, perhaps even at top of the file

src/WeaverQuery.coffee
```diff
@@ -109,21 +109,29 @@ class WeaverQuery
...
+    if _.isString(node)
+      node
+    else if node instanceof Weaver.Node
```
malamili: Does this break in case of BBClasses?

malamili: Perhaps check if there is a function .id() is better

test/WeaverQuery.test.coffee
```diff
@@ -241,6 +241,24 @@ describe 'WeaverQuery Test', ->
       )
     )
 
+  it 'should do relation hasRelationOut with subclasses', ->
```
malamili: Nice!

PR #156, opened 2017-10-10
Makes batchSave more batchy

gvdent:
Fixes saveBatch so the write operations from all nodes passed are collected and send as a batch instead of batches of individual saves.

### DIFF
--- a/src/CoreManager.coffee
+++ b/src/CoreManager.coffee
@@ -288,6 +288,24 @@ class CoreManager
       res
     )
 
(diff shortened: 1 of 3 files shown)


malamili:
Im worried that the batch can be too large. Should be send in splitted batches of max 500 ops each?

gvdent:
Agreed we'll need to configure batchsize somewhere eventually, ideally in the CoreManager so it applies not just to this but to all write operations being performed.

PR #157, opened 2017-10-11
Weav 184 ignores out of date

gvdent:
Adds support for traverseReplaces operation from weaverplatform/weaver-sdk-java#26


malamili:
Very nice!

PR #160, opened 2017-10-17
Operations will now be executed in batches of 500 (specified by Gijs …

mathieub93:
Operations will now be executed in batches of 500 (specified by Gijs as initial setting)

Tested through WeaverNode with operation arrays up to 9 and batches of 1-4.
No additional tests written

### DIFF
--- a/src/CoreManager.coffee
+++ b/src/CoreManager.coffee
@@ -67,13 +67,9 @@ class CoreManager
     )
 
   executeOperations: (allOperations, target) ->
-    if (allOperations.length <= @maxBatchSize)
-      @POST('write', {operations: allOperations}, target)
-    else
-      operations = allOperations.splice(0, @maxBatchSize)
-      @POST('write', {operations}, target).then ( =>
-        @executeOperations(allOperations, target)
-      )
+    Promise.mapSeries(_.chunk(allOperations, @maxBatchSize), (operations) =>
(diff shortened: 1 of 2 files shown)


src/CoreManager.coffee
```diff
@@ -65,9 +66,14 @@ class CoreManager
       localTime - serverTime
     )
 
-
-  executeOperations: (operations, target) ->
-    @POST('write', {operations}, target)
+  executeOperations: (allOperations, target) ->
```
malamili: Can you try this code instead with your tests? More concise, no code duplication and better readable / understandable (no if statement, no recursion),

```
Promise.mapSeries(_.chunk(allOperations, @maxBatchSize), (operations) =>
     @POST('write', {operations}, target)
)
```

gvdent: One thing: does this new code maintain the order of operations over multiple batches? E.g. if batch1 creates a node, and in batch2 a relations is made to that node, it is guaranteed that batch1 is always executed before batch2?

malamili: Yes it is guaranteed, _.chunk works in order of elements and mapSeries ensures the promise is resolved before executing the next batch

mathieub93: This code works as intended, new PR will be made in a minute

PR #161, opened 2017-10-19
adding test to sanitize weaver query proposal

carvilsi:


test/WeaverQuery.test.coffee
```diff
@@ -1011,4 +1011,18 @@ describe 'WeaverQuery Test', ->
     ).then(->
       Weaver.Node.load('someNode')
     )
-    
+
+  it 'should sanitize weaver query', ->
+
```
gvdent: I'd include some nodes in this tests to ensure no funny things happen. E.g.

```
a = new Weaver.Node()
a.set('name', 'testy')
a.save().then( ->
  new Weaver.Query()
  .equals("name", "test'y")
  .find()
).then((res) ->
  expect(res).to.have.length.be(0)
)
```
Otherwise queries may return unexpected results when potentially hazardous characters are used.

test/WeaverQuery.test.coffee
```diff
@@ -1011,4 +1011,18 @@ describe 'WeaverQuery Test', ->
...
+    new Weaver.Query()
+    .contains("name ' ","one 'two ' three")
```
gvdent: A description might actually contain a string like this... In fact a name as well. I'd much like to search a weaver-database for all mentions of my favourite Narn "G'Kar".

gvdent:
Looks good to me, mergeback is good pending a passing build

PR #162, opened 2017-10-20
Weav 203 profile weaverquery end to end

mathieub93:
Times are being saved in the profile object.

Test time has increased from 15 seconds to 1 minute -> shouldn't be merged like this.


src/SocketController.coffee
```diff
@@ -29,13 +29,17 @@ class SocketController
...
   emit: (key, body) ->
+    emitStart = Date.now()
     new Promise((resolve, reject) =>
       @io.emit(key, JSON.stringify(body), (response) ->
         if response.code? and response.message?
           reject(new Error(response.message, response.code))
         else if response is 0
           resolve()
         else
+          emitEnd = Date.now()
```
malamili: Why only on the else? Why not also when rejected?

malamili:
I am missing Weaver.Query changes, I want to see the times in the Weaver.Query.profile() message, now the times are lost in the response.body

src/SocketController.coffee
malamili: this time it the full roundtrip, I want a split between the time from client to server, and server back to client

src/SocketController.coffee
```diff
@@ -29,17 +29,30 @@ class SocketController
     )
 
   emit: (key, body) ->
+    emitStart = Date.now()
     new Promise((resolve, reject) =>
-      @io.emit(key, JSON.stringify(body), (response) ->
+      @io.emit(key, JSON.stringify(body), (response) =>
+        response.sdkStart = emitStart
```
malamili: Why set this on the response? Looks hacky to me, just pass these times as arguments to the calculateTimestamps function

malamili: emitStart is defined outside this function which is a callback, how sure are you that it is guaranteed that the same time is passed? Not 100%, but I think this is broken..

src/SocketController.coffee
```diff
@@ -29,17 +29,30 @@ class SocketController
...
+        emitEnd = Date.now()
+        response.sdkEnd = emitEnd
+        response.totalTime = (emitEnd-emitStart)
+        @calculateTimestamps(response)
       )
     )
 
+  calculateTimestamps: (response) ->
+    response.sdkToServer  = response.serverStart - response.sdkStart
+    response.serverToConn = response.executionTimeStart - response.serverStart
```
malamili: executionTimeStart is already way ahead in the connector, its better to get time at entry point in the connector (somewhere at the route beginning)

src/SocketController.coffee
mathieub93: This works as intended but will change it up with the previous comment

mathieub93: Response is the object which is given from the connector back to the Sdk and has all the timestamps which are required to know the time. Adding these times to the response isn't required but seemed like the nicest implementation to keep data together.

Will change this.

src/SocketController.coffee
mathieub93: If you'd do this you would have to check the item between the start of the connector and the start of the execution stopwatch aswel. Otherwise the sum of times doesnt add up to equal the total time.
Also it won't be possible to check times by using the stopwatch itself.
This is not changed (I overlooked it) in the new PR

malamili: Hmm lets discuss this in person tomorrow

PR #163, opened 2017-10-20
Fixes the authorization test that fails when its ran on the server fo…

gvdent:
…r the not-first time

Fix for @carvilsi


carvilsi:
Perfect! Thanks @gvdent 👍

PR #164, opened 2017-10-20
removing ancient project parameter for weaverNode class

carvilsi:


carvilsi:
👍 
In case we'll want to have this use case, save one set to one project and then other set to other project, we'll need to add this functionality at the server side. That in the current state is not checking for this project parameter at the write function.

malamili:
Please dont apply this, we need this very shortly in the navigator where we'll give bulk operations accross projects..the client will then supply the project id

gvdent:
See @malamili's comment, closing this

PR #165, opened 2017-10-20
Override ignore update

carvilsi:
- Handling nicely errors on update out of date
- Feature to override default behaviour when `weaver.setOptions({ignoresOutOfDate: true})`

### DIFF
--- a/src/WeaverNode.coffee
+++ b/src/WeaverNode.coffee
@@ -107,11 +107,11 @@ class WeaverNode
 
 
 
-  set: (field, value) ->
+  set: (field, value, ignoresOutOfDate) ->
     if field is 'id'
       throw Error("Attribute 'id' cannot be set or updated")
 
-    # Get attribute datatype, TODO: Support date
+    # Get attribute datatype
     dataType = null
     if util.isString(value)
       dataType = 'string'
@@ -130,7 +130,7 @@ class WeaverNode
         throw new Error("Specifiy which attribute to set, more than 1 found for " + field) # TODO: Support later
 
       oldAttribute = @attributes[field][0]
-      newAttributeOperation = Operation.Node(@).createAttribute(field, value, dataType, oldAttribute.nodeId, Weaver.getInstance()._ignoresOutOfDate)
+      newAttributeOperation = Operation.Node(@).createAttribute(field, value, dataType, oldAttribute.nodeId, Weaver.getInstance()._ignoresOutOfDate if !ignoresOutOfDate?)
     else
       newAttributeOperation = Operation.Node(@).createAttribute(field, value, dataType)
 
--- a/src/WeaverNode.coffee
+++ b/src/WeaverNode.coffee
@@ -275,6 +275,10 @@ class WeaverNode
         @
       ).catch((e) =>
 
+        console.log '=^^=|_THE ERROR ON SAVE________________________|=^^='
+        console.log e
+        console.log '=^^=|___________________THE ERROR ON SAVE_______|=^^='
+
         # Restore the pending writes to their originating nodes
         # (in reverse order so create-node is done before adding attributes)
         for i in writes by -1
--- a/src/WeaverNode.coffee
+++ b/src/WeaverNode.coffee
@@ -107,7 +107,7 @@ class WeaverNode
 
 
 
-  set: (field, value, ignoresOutOfDate) ->
+  set: (field, value, options) ->
     if field is 'id'
       throw Error("Attribute 'id' cannot be set or updated")
 
@@ -130,7 +130,7 @@ class WeaverNode
         throw new Error("Specifiy which attribute to set, more than 1 found for " + field) # TODO: Support later
 
       oldAttribute = @attributes[field][0]
-      newAttributeOperation = Operation.Node(@).createAttribute(field, value, dataType, oldAttribute.nodeId, Weaver.getInstance()._ignoresOutOfDate if !ignoresOutOfDate?)
+      newAttributeOperation = Operation.Node(@).createAttribute(field, value, dataType, oldAttribute.nodeId, Weaver.getInstance()._ignoresOutOfDate if !options?.ignoresOutOfDate?)
     else
       newAttributeOperation = Operation.Node(@).createAttribute(field, value, dataType)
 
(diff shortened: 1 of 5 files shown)


src/WeaverNode.coffee
```diff
@@ -107,11 +107,11 @@ class WeaverNode
 
 
 
-  set: (field, value) ->
+  set: (field, value, ignoresOutOfDate) ->
```
malamili: if we will add more options here, we'll quickly break backwards compatibility, better to use an option object here...adding options later will not break

src/WeaverNode.coffee
```diff
@@ -275,6 +275,15 @@ class WeaverNode
         @
       ).catch((e) =>
 
+        err = String(e)
```
malamili: This error parsing is too failure sensitive, this should move to the connector, catch the correct Postgres exception and also supply a bit more information such as old value / id etc. Also add a code, in the client we dont want to test for a message (in the increment function later on)

test/WeaverNode.test.coffee
```diff
@@ -570,15 +570,15 @@ describe 'WeaverNode test', ->
...
     a.set('name', 'first')
     alsoA = undefined
-
+    options = {ignoresOutOfDate: true}
     a.save().then(->
       Weaver.Node.load(a.id())
     ).then((node) ->
       alsoA = node
-      a.set('name', 'second', false)    # checking for the existence of the ignoresOutOfDate parameter so any value passed here will overrides the {ignoresOutOfDate: true} state
+      a.set('name', 'second', options)    # checking for the existence of the ignoresOutOfDate parameter so any value passed here will overrides the {ignoresOutOfDate: true} state
```
mathieub93: It's not checking for the true/false value anymore but for the key/value pair within the object. Besides this, it looks good

carvilsi: We where just checking the existence of the parameter not the value, but we changed this to be inside an object trying to be more maintainable due to version compatibility.

test/WeaverNode.test.coffee
```diff
@@ -532,3 +540,49 @@ describe 'WeaverNode test', ->
...
+
+  it 'should fail trying to save a node with the same id than the destroyed node', ->
+    a = new Weaver.Node('theid')
+    a.save()
+    .then( ->
+      a.destroy()
+    ).then( ->
+      new Weaver.Node('theid').save()
+    ).catch((err) ->
+      assert.equal(err,'Error: The id theid already exists')
```
gvdent: This test also passes if no rejection occurs, instead use: 
```
a.save()
.then(..)
.then(...)
.should.be.rejectedWith('Error: the id theid already exists')
```

test/WeaverNode.test.coffee
```diff
@@ -532,3 +540,49 @@ describe 'WeaverNode test', ->
...
+  it 'should fail trying to save a node saved with another attribute value', ->
+    a = new Weaver.Node('theid')
+    a.set('name','Toshio').save()
+    .then( ->
+      Weaver.Node.load('theid')
+    ).then((loadedNode) ->
+      loadedNode.set('name','Samantha')
+      loadedNode.save()
+    ).catch((err) ->
+      assert.equal(err,'Error: The id theid already exists')
```
gvdent: See other rejectedWith comment.

test/WeaverNode.test.coffee
```diff
@@ -532,3 +540,49 @@ describe 'WeaverNode test', ->
...
+      Weaver.Node.load(a.id())
+    ).then((node) ->
+      alsoA = node
+      a.set('name', 'second', options)    # checking for the existence of the ignoresOutOfDate parameter so any value passed here will overrides the {ignoresOutOfDate: true} state
+      a.save()
+    ).then(->
+      alsoA.set('name', 'allegedly updates first', options)
+      alsoA.save()
+    ).catch((err) ->
+      assert.equal(err,"Error: The attribute that you are trying to update is out of synchronization with the database, therefore it wasn\'t saved")
```
gvdent: See other rejectedWith comment.

test/WeaverNode.test.coffee
```diff
@@ -221,7 +223,7 @@ describe 'WeaverNode test', ->
     ).then(->
       assert(false)
     ).catch((error) ->
-      assert.equal(error.code, Weaver.Error.NODE_ALREADY_EXISTS)
+      assert.equal(error, "Error: The id double-node already exists")
```
gvdent: Please use the chai-as-promised structure here, so `.should.be.rejectedWith(...)` instead of then assertFalse and checking the error.

test/WeaverNode.test.coffee
```diff
@@ -418,7 +420,9 @@ describe 'WeaverNode test', ->
         ay.save(),
         aay.save()
       ])
-    ).should.eventually.be.rejected
+    ).catch((err) ->
+      assert.equal(err,"Error: The attribute that you are trying to update is out of synchronization with the database, therefore it wasn\'t saved")
```
gvdent: See other rejectedWith comment.

test/WeaverNode.test.coffee
```diff
@@ -471,7 +475,9 @@ describe 'WeaverNode test', ->
     ).then(->
       alsoA.set('name', 'allegedly updates first')
       alsoA.save()
-    ).should.eventually.be.rejected
+    ).catch((err) ->
+      assert.equal(err,"Error: The attribute that you are trying to update is out of synchronization with the database, therefore it wasn\'t saved")
```
gvdent: See other rejectedWith comment.

test/WeaverNode.test.coffee
```diff
@@ -509,8 +515,10 @@ describe 'WeaverNode test', ->
     ).then(->
       alsoA.relation('rel').update(b, d)
       alsoA.save()
-    ).should.eventually.be.rejected
-  
+    ).catch((err) ->
+      assert.equal(err,"Error: The relation that you are trying to update is out of synchronization with the database, therefore it wasn\'t saved")
```
gvdent: See other rejectedWith comment.

test/WeaverNode.test.coffee
```diff
@@ -532,3 +540,49 @@ describe 'WeaverNode test', ->
...
+    a.save()
+    .then( ->
+      a.destroy()
+    ).then( ->
+      new Weaver.Node('theid').save()
+    ).catch((err) ->
+      assert.equal(err,'Error: The id theid already exists')
+    )
+
+  it 'should fail trying to save a node saved with another attribute value', ->
```
gvdent: This test does not make too much sense to me currently. Is this current functionality? Because this looks like just a regular attribute update or attribute addition.

test/WeaverNode.test.coffee
```diff
@@ -532,3 +540,49 @@ describe 'WeaverNode test', ->
...
+      Weaver.Node.load('theid')
+    ).then((loadedNode) ->
+      loadedNode.set('name','Samantha')
+      loadedNode.save()
+    ).catch((err) ->
+      assert.equal(err,'Error: The id theid already exists')
+    )
+
+
+  it 'should allow to override the out-of-sync attribute updates at the set operation if the ignoresOutOfDate flag is set', ->
```
gvdent: Should ignoresOutOfDate in the individuals save options here not be`false`?

carvilsi: Like we've been talking we are just checking for the existence of this options value.

PR #170, opened 2017-10-26
adding batchDestroy and tests for this functionallity

carvilsi:


test/WeaverQuery.test.coffee
```diff
@@ -1001,7 +1001,7 @@ describe 'WeaverQuery Test', ->
         expect(queryResult.nodes[0].nodeId).to.equal('someNode')
       )
 
-    it 'should know all timestamps and have them logically correct', ->
+    it.skip 'should know all timestamps and have them logically correct', ->
```
gvdent: This test should not be disabled in this PR. If the test passes on develop it should not be broken in this feature. 

Previously the build did not pass on develop but that change has been reverted. Once most recent changes are merged into this branch this change should not longer be required.

test/WeaverNode.test.coffee
```diff
@@ -591,3 +591,52 @@ describe 'WeaverNode test', ->
...
+    )
+
+  it 'should handler an error when trying batch delete nodes with any node', ->
+
+    Weaver.Node.batchDestroy()
+    .catch((error)->
+      assert.equal(error, "Cannot batch destroy nodes without any node")
+    )
+
+  it 'should reject when trying batch delete nodes without proper nodes', ->
```
gvdent: Nice, I like the edge case testing :)

malamili:
What is the performance on this batchDestroy? The initial idea was to let the connector implement batch, because im afraid this is still slow (every destroy does a lookup).
Could you test a batch destroy of 1000nodes and report the time?

carvilsi:
The speed test result for 1000 nodes is:
**batchSave:** 206.731ms
**batchDestroy:** 38862.601ms

carvilsi:
Merging right now and leaving the improving task for batch destroy nodes at weaver-postgrSQL-connector for the future.

PR #171, opened 2017-10-27
Fixes develop build by using the correct connector version

gvdent:


gvdent:
Disregard that, doesn't fix it.

PR #174, opened 2017-10-27
Weav 210 implement sdk increment feature after 206 and 207

mathieub93:
Increment function now takes out-of-sync field to be updated 'on top of eachother'
Added a generic isOutOfDate function which is being used in the increment function


test/WeaverNode.test.coffee
```diff
@@ -170,6 +170,25 @@ describe 'WeaverNode test', ->
       assert.equal(loadedNode.get('length'), 5)
     )
 
+  it 'should increment an existing out-of-sync number attribute', ->
```
malamili: good test!

src/WeaverNode.coffee
```diff
@@ -107,7 +107,7 @@ class WeaverNode
 
 
 
-  set: (field, value) ->
+  set: (field, value, options) ->
```
malamili: Wasnt this options parameter already from another branch present?

src/WeaverNode.coffee
```diff
@@ -149,22 +149,39 @@ class WeaverNode
...
   increment: (field, value, project) ->
-
     if not @attributes[field]?
       throw new Error("There is no field " + field + " to increment")
     if typeof value isnt 'number'
       throw new Error("Field " + field + " is not a number")
 
-    currentValue = @get(field)
-    @set(field, currentValue + value)
+    @isOutofDate(field).then((outOfDate) =>
```
malamili: I dont like this implementation, this means on every increment, we do a full Weaver.Node.load, which becomes slow when done a lot. Even worse, you're doing a Weaver.Node.load twice! First to check if outOfDate, then again when indeed outOfDate. In general, try loading as little as possible. 

The point of previous issues 206 and 207 was that you can try to just save and catch the error if it fails..then on the fail get the real value and update with that one. This way you only load when it fails.

src/WeaverNode.coffee
```diff
@@ -149,22 +149,39 @@ class WeaverNode
...
+      if outOfDate
+        Weaver.Node.load(@nodeId).then((loadedNode) =>
+          currentValue = loadedNode.get(field)
+          @set(field, currentValue + value)
+        ).then(=>
```
malamili: this then is unnecessary, @set does not return a promise.

src/WeaverNode.coffee
```diff
@@ -149,22 +149,39 @@ class WeaverNode
...
+            currentValue + value
+          )
+        )
 
-    # To be backwards compatible, but its better not to save here
-    @save().then(->
-      # Return the incremented value
-      currentValue + value
+      else if !outOfDate
+        currentValue = @get(field)
```
malamili: These lines are duplicated before, try to be DRY

src/WeaverNode.coffee
mathieub93: I believe this was implemented in 206 (and ready to merge) and since Carlos is back today, I will take a look at this. so I don't duplicate the change.
This was mentioned in the commit aswel to have duplicate code.

PR #180, opened 2017-11-01
Weav 210 implement sdk increment feature

mathieub93:
Working feature
This feature is based on Carlos his fork and therefor I will work with Carlos to merge this as soon as 206 and 207 PR's are accepted along with this one.

### DIFF
--- a/src/WeaverNode.coffee
+++ b/src/WeaverNode.coffee
@@ -185,7 +185,7 @@ class WeaverNode
         pendingNewValue
       ).catch((error) =>
         if (error.message == 'The attribute that you are trying to update is out of synchronization with the database, therefore it wasn\'t saved')
-          # @pendingWrites.splice(-1, 1)
+          @pendingWrites.splice(-1, 1) # remove last (failing) operation, otherwise the save() keeps on failing on this node
           @increment(field, value, project, true)
       )
 
--- a/src/WeaverNode.coffee
+++ b/src/WeaverNode.coffee
@@ -152,50 +152,49 @@ class WeaverNode
 
 
   # Update attribute by incrementing the value, the result depends on concurrent requests, so check the result
-  increment: (field, value = 1, project, outOfSync = false) ->
+  increment: (field, value = 1, project) ->
 
     if not @attributes[field]?
       throw new Error("There is no field " + field + " to increment")
     if typeof value isnt 'number'
       throw new Error("Field " + field + " is not a number")
 
-    if (outOfSync)
-      new Weaver.Query()
-      .select(field)
-      .restrict(@id())
-      .first()
-      .then((loadedNode) =>
-        currentValue = loadedNode.get(field)
-        pendingNewValue = currentValue + value
-        loadedNode.set(field, pendingNewValue)
+    currentValue = @get(field)
+    pendingNewValue = currentValue + value
+    @set(field, pendingNewValue)
 
-        # To be backwards compatible, but its better not to save here
-        loadedNode.save().then(->
-          # Return the incremented value
-          pendingNewValue
-        )
-      )
-    else
-      currentValue = @get(field)
+    # To be backwards compatible, but its better not to save here
+    @save().then(=>
+      # Return the incremented value
+      pendingNewValue
+    ).catch((error) =>
+      switch error.code
+        when WeaverError.WRITE_OPERATION_INVALID then return @_incrementOfOutSync(field, value, project)
+
+      # Will be removed when switch above does recognise error.code
+      if (error.message == 'The attribute that you are trying to update is out of synchronization with the database, therefore it wasn\'t saved')
+        index = @pendingWrites.map((o) => o.key).indexOf(field) # find failed operation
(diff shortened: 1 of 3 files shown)


test/WeaverNode.test.coffee
```diff
@@ -170,6 +170,28 @@ describe 'WeaverNode test', ->
...
+    node.save().then(->
+      Weaver.Node.load(node.id())
+    ).then((loadedNode) ->
+      assert.equal(loadedNode.get('length'), 3)
+      sameNode = loadedNode
+      node.increment('length', 4)
+    ).then((value) ->
+      assert.equal(value, 7)
+      sameNode.increment('length', 5)
+    ).catch((err) ->
```
malamili: This should be thrown to the client, only internally

src/WeaverNode.coffee
```diff
@@ -151,21 +151,43 @@ class WeaverNode
...
+      pendingNewValue = currentValue + value
+      @set(field, pendingNewValue)
+
+      # To be backwards compatible, but its better not to save here
+      @save().then(=>
+        # Return the incremented value
+        pendingNewValue
+      ).catch((error) =>
+        if (error.message == 'The attribute that you are trying to update is out of synchronization with the database, therefore it wasn\'t saved')
+          # @pendingWrites.splice(-1, 1)
```
malamili: How is this not needed?

src/WeaverNode.coffee
```diff
@@ -151,21 +151,43 @@ class WeaverNode
...
+      pendingNewValue = currentValue + value
+      @set(field, pendingNewValue)
+
+      # To be backwards compatible, but its better not to save here
+      @save().then(=>
+        # Return the incremented value
+        pendingNewValue
+      ).catch((error) =>
+        if (error.message == 'The attribute that you are trying to update is out of synchronization with the database, therefore it wasn\'t saved')
+          @pendingWrites.splice(-1, 1) # remove last (failing) operation, otherwise the save() keeps on failing on this node
```
gvdent: You can't guarantee that the last pending write is the increment (although it often will be). 

Since in between the save() and the reject additional write operations may have been added and the failing write operations are re-added at the start of the pending operations. 

Iterate over the pendingWrites and remove the attribute update for the field being updated, as that guaranteed to be the failing operation.

src/WeaverNode.coffee
```diff
@@ -151,21 +151,43 @@ class WeaverNode
 
 
   # Update attribute by incrementing the value, the result depends on concurrent requests, so check the result
-  increment: (field, value, project) ->
+  increment: (field, value = 1, project, outOfSync = false) ->
```
gvdent: Style issue:
I'd move the entire outOfSync = true block to a seperate _incrementOutOfSync function (the _ to indicate a private function as is the convention). One reason is that besides the argument checking (kudos for that btw) the function has two different implementations depending on a boolean argument, which is a code smell. The second reason is that this is an argument to a public function which an SDK user should never even think about, it's an implementation detail.

src/WeaverNode.coffee
```diff
@@ -151,21 +151,43 @@ class WeaverNode
...
+      currentValue = @get(field)
+      pendingNewValue = currentValue + value
+      @set(field, pendingNewValue)
+
+      # To be backwards compatible, but its better not to save here
+      @save().then(=>
+        # Return the incremented value
+        pendingNewValue
+      ).catch((error) =>
+        if (error.message == 'The attribute that you are trying to update is out of synchronization with the database, therefore it wasn\'t saved')
```
gvdent: Please don't check on string literals for error messages. If the specific error message does not have its own code it should get that. Otherwise when an error message is changed, this functionality breaks. 

Also, this lacks an else statement which should Promise.reject(error) to ensure other errors are not missed by calling code.

src/WeaverNode.coffee
```diff
@@ -167,20 +168,44 @@ class WeaverNode
...
-    @save().then(->
+    @save().then(=>
       # Return the incremented value
-      currentValue + value
+      pendingNewValue
+    ).catch((error) =>
+      if (error.code == WeaverError.WRITE_OPERATION_INVALID)
+        index = @pendingWrites.map((o) => o.key).indexOf(field) # find failed operation
+        @pendingWrites.splice(index, 1) if index > -1 # remove failing operation, otherwise the save() keeps on failing on this node
+        @_incrementOfOutSync(field, value, project)
```
gvdent: This needs an 
```
else 
  Promise.reject()
```
Because now any errors that are not the WRITE_OPERATION_INVALID code are eaten up and the promise is resolved, which should not happen on an error.

src/WeaverNode.coffee
```diff
@@ -167,20 +168,44 @@ class WeaverNode
...
+    )
+
+  _incrementOfOutSync: (field, value, project) ->
+
+    new Weaver.Query()
+    .select(field)
```
gvdent: hah, nice! 👍

PR #191, opened 2017-11-14
Sanitize query

carvilsi:
Adding tests for weaver-query sanitize.

### DIFF
--- a/test/WeaverQuery.test.coffee
+++ b/test/WeaverQuery.test.coffee
@@ -823,6 +823,52 @@ describe 'WeaverQuery Test', ->
         )
       )
 
+
+  describe 'handling SQL injection and other parsing things', ->
+
+    it 'should sanitize weaver query', ->
+
+      a = new Weaver.Node()
+      a.set('name',"test")
+      a.save().then( ->
+        new Weaver.Query()
+        .contains("name","test'y")
+        .hasRelationOut("' link")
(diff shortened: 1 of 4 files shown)


test/WeaverQuery.test.coffee
```diff
@@ -823,6 +823,52 @@ describe 'WeaverQuery Test', ->
...
+    it 'should sanitize weaver query with simple quote on restrict', ->
+
+      a = new Weaver.Node('node_id')
+      a.set('name',"test")
+      a.save().then( ->
+        new Weaver.Query()
+        .restrict("'node_id'")
+        .find()
+      ).then((nodes)->
+        expect(nodes.length).to.equal(1)
```
gvdent: I'd expect nodes to have length 0 here, since there is no node with id `'node_id'`, you only create one with id `node_id`, and those are not equal.

mathieub93: Tested this on some random test and the .then((nodes) -> is never executed and therefor the test is not failing.

gvdent: `new Weaver.Query().restrict(['totallydoesnotexist']).find().then(console.log)` gives me an empty array... If the then is never executed there was some other error (test with .catch(console.error) or the likes).

gvdent:
Closing PRs that have not been updated for a month.

PR #193, opened 2017-11-16
Weav 217 implement unrecoverable delete

mathieub93:


test/WeaverNode.test.coffee
```diff
@@ -769,3 +769,74 @@ describe 'WeaverNode test', ->
...
+    id = node1.id()
+    node1.save().then(->
+      node1.destroy()
+    ).then(->
+      node2.save()
+    ).then(->
+      Weaver.Node.load(node2.id())
+    ).then((loadedNode) ->
+      assert.equal(loadedNode.id(), id)
+      weaver.setOptions({unrecoverableRemove: false})
```
gvdent: Restoring the options should be done in a .finally otherwise if the test fails subsequent test may also fail.

src/WeaverNode.coffee
```diff
@@ -350,18 +350,29 @@ class WeaverNode
...
+  # Removes node, with the option to remove it unrecoverable
   destroy: (project) ->
     cm = Weaver.getCoreManager()
     cm.enqueue( =>
-      if @nodeId?
-        cm.executeOperations([Operation.Node(@).removeNode()], project).then(=>
-          Weaver.publish('node.destroyed', @id())
-          delete @[key] for key of @
+
+      if (Weaver.getInstance()._unrecoverableRemove)
```
gvdent: I think this should also be an option on the operation in case it's more of a one-time-only permanent delete instead of an always forever thing. 

Currently you'd need to set the global unrecoverableRemove option, which may then apply to operations where you may actually want to be able to recover.

PR #196, opened 2017-11-20
Weav 216 check existence of node in one query

mathieub93:
In the end I only had to write one function which uses the already existent restrict which creates a singular query based on a set of nodes/node ids.
Tests have been added for small and large amounts of nodes.


test/WeaverQuery.test.coffee
```diff
@@ -1050,3 +1050,48 @@ describe 'WeaverQuery Test', ->
...
+        expect(result.c).to.be.false;       expect(result.d).to.be.true
+        expect(result.e).to.be.false;       expect(result.f).to.be.false
+        expect(result.g).to.be.true;        expect(result.h).to.be.false
+        expect(result.i).to.be.false;       expect(result.j).to.be.true
+        expect(result.k).to.be.true;        expect(result.l).to.be.true
+        expect(result.m).to.be.true
+      )
+    )
+
+  it 'should be able to check existance on a list of many Weaver nodes', ->
```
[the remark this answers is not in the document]
malamili: agree

src/WeaverQuery.coffee
```diff
@@ -271,6 +271,30 @@ class WeaverQuery
...
+    @restrict(nodes)
+    @find().then((results)->
+      sortedNodes = _.sortBy(nodes, ['nodeId']);
+      sortedResults = _.sortBy(results, ['nodeId']);
+
+      #First set all nodes to false, follow loops will only check for true values
+      for node in sortedNodes
+        map[node.id()] = false
+
+      # Algorithm to find all existing nodes, twice as fast as nested for loop on 10000 nodes.
```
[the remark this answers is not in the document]
malamili: Indeed nice! This however downloads attributes and (shallow) relations that are useless. For attributes, we can select none by setting @_select = [] i think. For the relations, im not sure how to not query these, if not possible, please create a JIRA ticket

src/WeaverQuery.coffee
```diff
@@ -282,6 +279,50 @@ class WeaverQuery
...
+      @useConstructorFunction = -> Constructor
+    Weaver.getCoreManager().findExistingNodes(@).then((results) =>
+
+      sortedNodes   = _.sortBy(nodes, ['nodeId'])
+      sortedResults = [].slice.call(results).sort()
+
+      @compareSortedNodeLists(sortedNodes, sortedResults)
+    )
+
+  compareSortedNodeLists: (nodes, compare) ->
```
gvdent: This can be private.

src/WeaverQuery.coffee
```diff
@@ -282,6 +279,50 @@ class WeaverQuery
...
+      sortedNodes = _.sortBy(nodes, ['nodeId']);
+      sortedResults = _.sortBy(results, ['nodeId']);
+
+      @compareSortedNodeLists(sortedNodes, sortedResults)
+    )
+
+  findExistingNodesInGraph: (nodes, graph) ->
+    map = {}
+    @restrict(nodes)
+    @restrictGraphs(graph or @collectGraphs(nodes))
```
gvdent: This doesn't use query functionality anymore, so the restrict/restrictGraphs should not be used anymore

src/WeaverQuery.coffee
```diff
@@ -282,6 +279,50 @@ class WeaverQuery
...
+
+    if Constructor?
```
gvdent: What is Constructor?

src/WeaverQuery.coffee
```diff
@@ -282,6 +279,50 @@ class WeaverQuery
     @_selectOut.push(relationKeys)
     @
 
+  findExistingNodes: (nodes) ->
```
gvdent: I don't think there should be two different functions here, these should be the same.

src/WeaverQuery.coffee
```diff
@@ -302,6 +302,47 @@ class WeaverQuery
...
+      if n[1] == ''
+        n[1] = undefined
+      newList.push(new Weaver.Node(n[0], n[1]))
+    newList
+
+  compareSortedNodeLists = (nodes, compare) =>
+    map = {}
+    #First set all nodes to false, follow loops will only check for true values
+    for node in nodes
+      idAndGraph = node.id() + "," + node.getGraph()
```
malamili: Could you please use the full node in the map? Instead of doing node,graph.
This makes it easier to parse for end users

test/test-suite.coffee
```diff
@@ -24,7 +24,7 @@ signInAsAdmin = ->
 
 beforeEach ->
   signInAsAdmin()
-    
+
```
gvdent: Most important change ;)

PR #205, opened 2017-11-24
Weav 229 add graph option to weaver sdk js

mathieub93:
Working feature, however I feel like the code is written less pretty then we would like it..


src/Operation.coffee
```diff
@@ -68,10 +74,13 @@ NodeOperation = (node) ->
...
       sourceId: node.id()
       key
-      targetId: to
-      replacesId: replaces
+      targetId: to.id()
+      replacesId: replacesId
       replaceId
       traverseReplaces: ignoreConcurrentReplace if replaces? and ignoreConcurrentReplace?
+      graph: node.getGraph()
+      targetGraph: to.getGraph()
```
malamili: Why do we make a targetGraph write op field? We cant change this right? Same is true for sourcegraph on attributes

malamili:
I think the code is nice, what aspect did you find not pretty? I do have a question about sourcegraph / targetgraph in the wripteops though, perhaps @gvdent can explain too.

test/WeaverNode.test.coffee
```diff
@@ -804,21 +804,6 @@ describe 'WeaverNode test', ->
       weaver.setOptions({unrecoverableRemove: false})
     )
 
-  it 'should be possible to get write operations from a node when weaver is not instantiated', ->
```
gvdent: Why would this no longer work?

src/Operation.coffee
gvdent: Every operation that uses an id now also needs a graph. So for every xId there should also be an xGraph

test/WeaverNode.test.coffee
mathieub93: Not sure why this is removed, added it and it's working fine.

src/WeaverNode.coffee
```diff
@@ -8,9 +8,10 @@ WeaverError = require('./WeaverError')
 
 class WeaverNode
 
-  constructor: (@nodeId) ->
+  constructor: (@nodeId, graph) ->
```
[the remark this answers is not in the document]
mathieub93: does this mean, that the line: @graph = graph if graph? can be completely removed aswel?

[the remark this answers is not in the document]
gvdent: yup :)

src/Operation.coffee
```diff
@@ -30,23 +34,41 @@ NodeOperation = (node) ->
...
+      replacesGraph: replaces.graph if replaces?
+
+      # Node of the new attribute
       replaceId
+      replaceGraph: graph if graph?
+
       traverseReplaces: ignoreConcurrentReplace if replaces? and ignoreConcurrentReplace?
+
+      # New attribute
+      replacedGraph: graph if graph?
```
gvdent: replacedGraph does not exist on the connector side, it can be removed (see `graph` comment)

src/Operation.coffee
```diff
@@ -30,23 +34,41 @@ NodeOperation = (node) ->
...
     replaceId = null
     replaceId = cuid() if replaces?
-
     {
       timestamp
       action: "create-attribute"
+
+      # This object (attribute)
       id: cuid()
+      graph: node.getGraph()
```
gvdent: This should be set to the graph argument with a default of node.getGraph() if no argument is provided.

src/Operation.coffee
```diff
@@ -56,21 +78,30 @@ NodeOperation = (node) ->
...
+    replacesId = replaces if util.isString(replaces) if replaces?
     replaceId = null
     replaceId = cuid() if replaces?
     throw new Error("Unable to set relation #{key} from #{node.id()} to null node") if !to?
+    throw new Error("Unable to set relation #{key} from #{node.id()} to null node") if !to.nodeId?
     {
       timestamp
       action: "create-relation"
       id
+      graph: node.getGraph()
```
gvdent: this should be `graph: graph or node.getGraph()` similar to the createAttribute function (needs an additional graph argument).

src/Operation.coffee
```diff
@@ -30,23 +34,41 @@ NodeOperation = (node) ->
       action: "remove-node-unrecoverable"
       id: node.id()
       removeId: cuid()
```
gvdent: This also needs a removeGraph for consistency (node.getGraph() is fine).

src/Operation.coffee
```diff
@@ -81,6 +112,7 @@ NodeOperation = (node) ->
       action: "remove-relation"
       id
       removeId: cuid()
+      graph: node.getGraph()
```
gvdent: - graph should be an argument since the id can be customized the graph should also be
- it needs a removeGraph (again defaulting to `node.getGraph()`)

src/Operation.coffee
mathieub93: Removing this will break the following test, is the test 'wrong' or should we keep replacedGraph?

    it 'should add create and remove statements to pendingWrites with graphs', ->
      node = new Weaver.Node('node1', 'first-graph')
      target = new Weaver.Node(null, 'second-graph')
      target2 = new Weaver.Node(null, 'third-graph')
      node.relation('link').add(target)
      node.relation('link').add(target2)
      node.set('age', 41, 'double', null, 'fourth-graph')
      node.set('age', 42, 'double', null, 'fifth-graph')
      node.destroy()
      expect(node.pendingWrites[0].graph).to.equal('first-graph')
      expect(target.pendingWrites[0].graph).to.equal('second-graph')
      expect(target2.pendingWrites[0].graph).to.equal('third-graph')
      expect(node.pendingWrites[1].replacedGraph).to.equal('fourth-graph')
      expect(node.pendingWrites[2].replacedGraph).to.equal('fifth-graph')

gvdent: So good that this is being tested, makes the consequence of a change more clear.

But indeed, the test is wrong, I expect the last two lines to read:
```
  expect(node.pendingWrites[1].graph).to.equal('fourth-graph')
  expect(node.pendingWrites[2].graph).to.equal('fifth-graph')
```

And for the fun of it, I'd add a 
```
node.set('age', 43, 'double') 
...
expect(node.pendingWrites[3].graph).to.equal('first-graph')
``` 
To test the default graph value which should equal that of the node graph.

PR #206, opened 2017-11-25
Exposing attributes() and relations()

malamili:
…tions


test-server/docker-compose.yml
```diff
@@ -24,6 +24,8 @@ services:
     - MINIO_ACCESS_KEY=NYLEXGR6MF2IE99LZ4UE
     - MINIO_SECRET_KEY=CjMuTRwYPcneXnGac2aQH0J+EdYdehTW4Cw7bZGD
     command: server /data
+    ports:
+    - "9000:9000"
```
gvdent: This looks like an accidental commit, keep ports that are not required for testing functionality on expose only. Unnecessary port exposure breaks builds.

test/WeaverModel.test.coffee
```diff
@@ -167,3 +167,17 @@ describe 'WeaverModel test', ->
         b.relation("buildBy").add(p1)
         b.relation("buildBy").add(p2)
         b.save()
+
+      it 'should list attributes', ->
+        p1 = new model.Person()
+        p1.set("fullName", "Hola 1")
+
+        assert.equal(p1.attributes()['fullName'], 'Hola 1')
```
gvdent: This looks nice :) 

Please change the tests that currently use _attributes to use this function instead, since we're treating the SDK tests as documentation on how to interact with SDK, so _members should not be the way we teach SDK users to interact with it.

src/WeaverRelationNode.coffee
```diff
@@ -7,17 +7,10 @@ class WeaverRelationNode extends WeaverNode
 
   constructor: (@nodeId) ->
     throw new Error("Please always supply a relId when constructing WeaverRelationNode") if not @nodeId?
-    @_stored = false       # if true, available in database, local node can hold unsaved changes
-    @_loaded = false       # if true, all information from the database was localised on construction
-
-    # Store all attributes and relations in these objects
-    @attributes = {}
-    @relations  = {}
```
gvdent: Nice cleanup 👍

test-server/docker-compose.yml
malamili: ah damn, i needed this..because i sometimes run the test-server setup without weaver-server, and then I need my local weaver-server to connect to minio..how bad does this break the build? can we circumvent?

test/WeaverModel.test.coffee
malamili: agreed..changing it now

PR #207, opened 2017-11-25
Allow to configure the proto relation

malamili:
Allowing to change proto relation by model definition


malamili:
good point, will change it

gvdent:
ASSUMING DIRECT CONTROL

PR #211, opened 2017-12-04
Query optimize

gvdent:
Intergation test for query optimization


test/WeaverHistory.test.coffee
```diff
@@ -153,22 +153,16 @@ describe 'WeaverHistory test', ->
       )
     )
 
-  it.skip 'should not allow sql injection in queries', ->
```
malamili: Yeaay not skipping, man's never skipping

PR #215, opened 2017-12-07
fix for increment issue

mingushead:


gvdent:
In between the save() call and the then argument being called another attribute set may be done which would then become it parameter before it being set back to the original value.

`.then(->` is the point where node's asynchronicity kicks in, between a promise being created and a promise being resolved a lot of time can pass, but more importantly a lot of code can be executed while the promise waits for an async call to be returned. 

This short lived set and reset of the global ignoreOutOfSync prevents other operations from using the temporary value.

PR #217, opened 2017-12-12
Fixes model database keys being loaded

gvdent:
Now to actually merge.


gvdent:
I'm considering the previous comments approval, deal with it.

PR #221, opened 2017-12-14
Weav 240 clean up timestamps within response object

mathieub93:


test/WeaverQuery.test.coffee
```diff
@@ -1067,3 +1073,30 @@ describe 'WeaverQuery Test', ->
...
+        expect(qr.serverToSdk).to.be.undefined
+
+        done()
+      )
+
+      node = new Weaver.Node('someNode')
+      node.save().then(->
+        Weaver.Node.load('someNode')
+      )
+    ).then(->
```
gvdent: This second load should not be required.

mathieub93: Removed it for both this and another function where it's not required.

PR #223, opened 2017-12-15
Weav 247 fix has relation graphs

gvdent:
Fixes a bunch of graph query related things


gvdent:
Closing for now, relation nodes still have lookup based on relation node id only.

PR #232, opened 2018-01-17
Weav 260 load relation node

gvdent:
Fixes loading relation nodes for relations defined to be in a graph.


src/WeaverRelation.coffee
```diff
@@ -41,10 +41,10 @@ class WeaverRelation
   to: (node)->
     relNode = @_getRelationNodeForTarget(node)
     throw new Error("No relation to a node with this id: #{node.id()}") if not relNode?
-    Weaver.RelationNode.load(relNode.id(), null, Weaver.RelationNode, true)
+    Weaver.RelationNode.load(relNode.id(), null, Weaver.RelationNode, true, relNode.getGraph())
 
   all: ->
-    (node for key, node of @nodes)
+    @nodes
```
bastbijl: this is a breaking change

gvdent: No?

nodes is an array, so (node for key, node of [ 4, 5, 6 ]) results in [ 4, 5, 6 ]. The only possible issue is that the array is no longer a copy but the array itself. But if you're modifying that array and expecting things to work you're going to have a bad time.

PR #238, opened 2018-01-23
Adds load function to model classes

gvdent:
Adds checking of loading instances and .load convenience function.


Gielert:
Looks good! 👍

PR #240, opened 2018-01-24
remove duplicate code

bastbijl:
Removes duplicate code. Also does not do a Weaver.publish event. This was unused functionality.


src/WeaverNode.coffee
```diff
@@ -339,25 +339,7 @@ class WeaverNode
 
   # Save node and all values / relations and relation objects to server
   save: (project) ->
-    cm = Weaver.getCoreManager()
-    writes = @_collectPendingWrites()
-
-    cm.enqueue(=>
-      cm.executeOperations((_.omit(i, "__pendingOpNode") for i in writes), project).then(=>
-        Weaver.publish('node.saved', i.__pendingOpNode) for i in writes
```
Gielert: You only removed a publish here for when nodes are saved. There is still PubSub functionality in the SDK.

gvdent: @bastbijl I thought the publish was moved over to the batchsave? Don't see it there anymore.

batchSave not having the publish seems like a bug to me, the functionality should be exactly the same, except batchSave requiring less network roundtrips.

PR #241, opened 2018-01-25
Ws 186 model relation out

gvdent:
Init members are now available on the model in addition to only the classes. This allows for easier querying.

### DIFF
--- a/src/WeaverModelQuery.coffee
+++ b/src/WeaverModelQuery.coffee
@@ -16,6 +16,18 @@ class WeaverModelQuery extends Weaver.Query
         Weaver.Node
     )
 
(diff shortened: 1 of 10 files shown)


src/WeaverModelQuery.coffee
```diff
@@ -16,6 +16,18 @@ class WeaverModelQuery extends Weaver.Query
         Weaver.Node
     )
 
+  getNodeIdFromStringOrNode: (node) ->
```
bastbijl: does this actually take a string too?

bastbijl: nevermind, i checkt it myself:()

PR #243, opened 2018-01-26
adds some Weaver.Node helper methods

mingushead:
Some helper methods to avoid having to loop through all those `node.relation().all()` arrays

### DIFF
--- a/src/WeaverNode.coffee
+++ b/src/WeaverNode.coffee
@@ -113,6 +113,15 @@ class WeaverNode
   relations: ->
     @_relations
 
+  isRelationPresent: (key, objId)->
+    for node in @relation(key).all()
+      return true if node.id() is objId
+    false
+
+  getRelation: (key, objId)->
+    for node in @relation(key).all()
+      return node if node.id() is objId
+
   _getAttributeValue: (attribute) ->
     if attribute.dataType is 'date'
       return new Date(attribute.value)
--- a/src/WeaverNode.coffee
+++ b/src/WeaverNode.coffee
@@ -113,6 +113,9 @@ class WeaverNode
   relations: ->
     @_relations
 
+  relationNodes: (key)->
+    @relation(key).relationNodes
+
   isRelationPresent: (key, objId)->
     for node in @relation(key).all()
       return true if node.id() is objId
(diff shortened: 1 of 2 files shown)


src/WeaverNode.coffee
```diff
@@ -113,6 +113,18 @@ class WeaverNode
   relations: ->
     @_relations
 
+  relationNodes: (key)->
+    @relation(key).relationNodes
+
+  isRelationPresent: (key, objId)->
```
gvdent: The second argument should a be an (optionally shallow, e.g. Weaver.Node.get) node, an id is not enough since that id could be present in another graph resulting in unexpected behaviour.

gvdent: Also, this should just be `@getRelation(key, node)?` instead of having its own implementation.

mingushead: So, to clarify:

`relationNodes(key)` is intended to be a shortcut (and properly accessed version) of doing `relation(key).relationNodes`

`getRelation(key, node)` was rather intended to be a shortcut for doing `return node when node.id() is id for node in relation(key).nodes` ie return the _object_ for _subject -> pred -> obj_, or `undefined` if the triple _subject -> predicate -> object_ dosen't exist

mingushead: so re your first comment, I will make the graph-related changes, but I don't see how the proposed `relationNodes` is the same as `@getRelation(key,node)` 
(the first returns the relation nodes, the second returns the object to the relation, or undefined)

gvdent: My point is that `node.id() is id` is never a valid check for the identity of a node, instead do `node.equals(otherNode)`, which internally does `node.id() is othernode.id() and node.graph() is othernode.id()` or something the like.

gvdent: also my other point was that
```
isRelationPresent: (key, node) ->
  getRelation(key, node)?
``` should be the implementation to prevent code duplication.

mingushead: Aaaaaah sorry, understood. I thought you were talking about `relationNodes`.
Of course, thanks for spotting it.

gvdent:
Closing this as it hasn't been updated in 14 days, please resubmit after requested changes have been made.

PR #244, opened 2018-01-26
Necessary model deformation

bastbijl:
Add existing node als member of class
Change prototype field in model to member and inherit
@gvdent


gvdent:
The full sdk is not needed for anything except bower, which iirc is the only reason it is part of the repo.

src/WeaverModel.coffee
```diff
@@ -43,15 +43,19 @@ class WeaverModel
...
           .first()
 
       @[className].load = load(className)
 
-    @_graphName = "#{@definition.name}-#{@definition.version}"
+    @_graph = "#{@definition.name}-#{@definition.version}"
 
   getGraphName: ->
-    @_graphName
+    console.warn('Deprecated function WeaverModel.getGraphName() used. Use WeaverModel.getGraph().')
```
gvdent: Nice 👍

src/WeaverModelClass.coffee
```diff
@@ -4,27 +4,41 @@ Weaver      = require('./Weaver')
...
-    super(nodeId, @model.getGraphName())
+  @classId: ->
+    "#{@definition.name}:#{@className}"
+
+  @addMember: (node)->
+    if node instanceof WeaverModelClass
+      node.nodeRelation(@model.definition.member).addInGraph(@asNode(), node.getGraph())
+    else
+      node.relation(@model.definition.member).addInGraph(@asNode(), node.getGraph())
+    node.save()
```
gvdent: Why is the node.save() required here? No other node modifications call save themselves, so unless there is no way around it (increment?) do not do this.

In general functions should serve a single purpose, see https://www.developerfusion.com/article/137636/taking-the-single-responsibility-principle-seriously/.

src/WeaverModelClass.coffee
```diff
@@ -4,27 +4,41 @@ Weaver      = require('./Weaver')
...
+
+  @asNode: ->
+    Weaver.Node.getFromGraph(@classId(), @model.getGraph())
```
gvdent: Please note that this returns a node that is not loaded (e.g. will not have the attributes/members of this instance). I'd suggest renaming it to getNode to remain consistent across the API where all getNode like calls return shallow nodes.

src/WeaverModel.coffee
```diff
@@ -75,29 +86,48 @@ class WeaverModel
...
-          # This is a member that is also a class
-          itemNode = nodesToCreate["#{@definition.name}:#{itemName}"]
-          # So add the prototype definition
-          itemNode.relation(@definition.prototype).addInGraph(nodesToCreate["#{@definition.name}:#{className}"], @getGraphName)
-          @[className][itemName] = itemNode
-        else
-          node = new ModelClass("#{@definition.name}:#{itemName}")
-          @[className][itemName] = node
-          nodesToCreate["#{@definition.name}:#{itemName}"] = node
+    Weaver.Node.batchSave(node for id, node of nodesToCreate).then(=>
```
gvdent: The intermediate batchSave here is only required because of the addMember side-effect, removing the save() in that function would ensure this is not required.

src/WeaverModel.coffee
```diff
@@ -75,29 +86,48 @@ class WeaverModel
...
+            if @[itemName]?
+              # This is a member that is also a class
+              itemNode = nodesToCreate["#{@definition.name}:#{itemName}"]
+              reg = @[className]
+              promises.push(
+                ModelClass.addMember(itemNode)
+                .then(->
+                  ModelClass.load(itemNode)
+                ).then((loaded)->
+                  reg[itemName] = loaded
```
gvdent: This structure is a mess and I'm not a fan of the save()/load() combination here. Is there a way to do this without database roundtrips? And neater?

src/WeaverModel.coffee
```diff
@@ -63,10 +67,17 @@ class WeaverModel
...
     @definition.inherit or '_inherit'
 
+  getMemberKey: ->
+    @definition.member or '_member'
+
+  getPrototypeKey: ->
+    console.warn('Deprecated function WeaverModel.getPrototypeKey() used. Use WeaverModel.getMemberKey().')
+    @getMemberKey()
+
   bootstrap: ->
```
gvdent: This function is becoming a maintenance hazard, if adjustments need to be made anywhere from a week from now, it will require substantial effort to get started. Preferably split up the functionality.

src/WeaverModelQuery.coffee
```diff
@@ -7,38 +7,35 @@ class WeaverModelQuery extends Weaver.Query
   constructor: (@model = Weaver.currentModel(), target) ->
     super(target)
 
-    # Define constructor function
-    @useConstructor((node) =>
```
gvdent: I'm not entirely sure we want this behaviour. The move seems to imply that WeaverModelQuery now returns plain Weaver.Nodes if a node in the result is not of the class argument passed. If this is the case this is backward compatibility breaking change.

If the change is needed because you want to force a certain type when quering nodes (e.g. force a certain shared supertype of nodes returned), this should be a seperate new method.

src/WeaverModelQuery.coffee
```diff
@@ -7,38 +7,35 @@ class WeaverModelQuery extends Weaver.Query
...
-          graph: node.model.getGraphName()
+          graph: node.model.getGraph()
         }
       else
         throw err
 
-  getPrototypeKey: ->
-    @model.definition.prototype or '_prototype'
 
   class: (modelClass) ->
```
gvdent: I can't completely oversee the implications of the change here, but my gut says this break backward compatibility.

src/WeaverNode.coffee
```diff
@@ -65,7 +65,10 @@ class WeaverNode
 
         instance = new Constructor(relation.target.nodeId, relation.target.graph)
         instance._loadFromQuery(relation.target, constructorFunction, fullyLoaded)
-        @relation(key).add(instance, relation.nodeId, false, relation.graph)
+        if @ instanceof Weaver.ModelClass
```
gvdent: 🚨 code smell 🚨 

A superclass should never check whether it is an instance of a child class, in fact it should never even know it has a subclass. Instead move this functionality to an _addRelation method which the subclass overrides.

test/WeaverModelQuery.test.coffee
```diff
@@ -70,6 +70,7 @@ describe 'WeaverModelQuery test', ->
 
       it 'should do an equalTo WeaverModelQuery', ->
         new Weaver.ModelQuery()
+        .class(model.Person)
```
gvdent: This highlights a backward compatibility issue, it is now required to specify a class restriction when that may not be what is wanted.

If for instance I want to query all nodes that have a relation to a certain document, but those may be members of different classes, how is this functionality now provided?

test/WeaverModelQuery.test.coffee
```diff
@@ -79,13 +80,13 @@ describe 'WeaverModelQuery test', ->
           assert.equal(p.constructor, model.Person)
           assert.equal(p.id(), "personA")
           assert.equal(p.get('fullName'), 'Aby Delores')
-          assert.equal(p.relation('hasHead').first().constructor, model.Head)  # <- this fails currently
+          assert.equal(p.relation('hasHead').first().constructor, Weaver.Node)
```
gvdent: No, this relation target should actually be a head, since we know it has to be a head because of the model definition.

test/WeaverModelQuery.test.coffee
```diff
@@ -79,13 +80,13 @@ describe 'WeaverModelQuery test', ->
...
         )
 
       it 'should load model instances', ->
         new Weaver.ModelQuery()
           .restrict("personB")
-          .first().should.eventually.be.instanceOf(model.Person)
+          .first(model.Person).should.eventually.be.instanceOf(model.Person)
```
gvdent: This should not need to be specified.

src/WeaverModel.coffee
bastbijl: Split up which functionality? You mean more than one pullrequest with the points from this pull request?

src/WeaverModelClass.coffee
bastbijl: Ill do the rename to getNode.

bastbijl: Why would you want a not not-loaded node? It won't have attributes or outgoing relations anyway because it is a modelclass repres. node.

src/WeaverModelQuery.coffee
bastbijl: Not forcing, but we have now multi parenting. So you have to say upfront how you want it instantiated. There is a check.

True, but the way this in implemeted now seems to me the right behaviour.

It shouldn't break, because the constructing now depends on the class specification. True for other nodes though.

I dont know how to make this backward compatible for that part.

test/WeaverModelQuery.test.coffee
bastbijl: I dont know how to do that. Ideas?

src/WeaverModel.coffee
gvdent: Split up the _bootstrapClasses method, it now contains some clearly delineated  steps which should have their own method.

gvdent: Split up _bootstrapClasses, it is becoming too big a method and contains some steps which look like they are clearly delineated, so make that explicit by putting their functionality in seperate methods with a descriptive name.

src/WeaverModelClass.coffee
gvdent: To me it seems like this function serves no purpose currently. If you want to refer to the model node (as a relation target for instance) you can just use the modelclass instance itself. If you want to add non-model data, performing these operations on shallow nodes is a bad practice and should not be encouraged by the api.

src/WeaverModelQuery.coffee
gvdent: It is not hard to imagine modelqueries without a class specification if you don't care about the specific class (see https://github.com/weaverplatform/weaver-sdk-js/pull/244#discussion_r164660915). This will break this functionality. I'd leave in the existing means to instantiate and call a constructor (whichever one in the case of multi parents), and add a new query method (asClass?) to call useConstructor. Or even better, just call useConstrutor manually if you insist on forcing one.

test/WeaverModelQuery.test.coffee
gvdent: Don't force .class() usage. Add an asClass method maybe.

PR #246, opened 2018-01-30
Adds test and new dependency fixing the test

gvdent:
Fixes a bug in the [no]Relation(In|Out) queries where the restriction on each clause would apply to each clause instead of to their individual ones.


gvdent:
PR depends on functionality from https://github.com/weaverplatform/weaver-database-postgresql/pull/84

PR #288, opened 2018-04-26
Pending writes get lost

bastbijl:
When more then one Weaver.Node instance exists for the same node (eg. node id and graph) the pending writes for only one of those are collected.

This is fixed by not using the node id as representation to detect cycles but using the object compare

objA === objB

I certify that:
- [ ] The automated build passes on the branch this pull request is from
- [ ] My change is documented in the Changelist.md
  another developer
  forget
  target branch
- [x] This is a quality PR


gvdent:
This is a mess

PR #297, opened 2018-05-08
Destroy node delayed

bastbijl:
Allow destroying nodes without preforming the action imediate. Used for batch deletes.

I certify that:
  another developer
  forget
  target branch
- [x] This is a quality PR


test/WeaverNode.test.coffee
```diff
@@ -110,6 +110,23 @@ describe 'WeaverNode test', ->
...
+    Weaver.Node.batchSave([a,b,c])
+    .then((node) ->
+      a.destroyDelayed()
+      b.destroyDelayed()
+      c.destroyDelayed()
+      Weaver.Node.batchSave([a,b,c])
+    ).then(->
+      Weaver.Node.load(a).should.be.rejectedWith(Weaver.Error.NODE_NOT_FOUND)
+      Weaver.Node.load(b).should.be.rejectedWith(Weaver.Error.NODE_NOT_FOUND)
+      Weaver.Node.load(c).should.be.rejectedWith(Weaver.Error.NODE_NOT_FOUND)
```
gvdent: This last three promise block is an issue. Only the last promise (should.be.rejectedWith returns a promise) is returned and used for the test result, so if either of the other two fail it will not mark it as a test failure.

Something like the following would fix it:
```
Promise.all([
  Weaver.Node.load(a).should.be.rejectedWith(WeaverError.NODE_NOT_FOUND),
  Weaver.Node.load(b)...,
  Weave...
])
```

That way all checks are returned as a single promise and the test will work properly.

PR #310, opened 2018-06-07
Not mount default.coffee

bastbijl:
Only mount local.coffee in docker composition


carvilsi:
Why only add the local.coffee config file to the docker volume instead the default? 
Seems to be rare, the local.coffee configuration is for local things, and we use to have this out of the git workflow.

test-server/docker-compose.yml
```diff
@@ -34,5 +34,5 @@ services:
     ports:
     - "9487:9487"
     volumes:
-    - "./weaver-server-config:/usr/src/app/config"
+    - "./weaver-server-config/local.coffee:/usr/src/app/config/local.coffee"
```
gvdent: Perfect :+1:

gvdent:
@carvilsi The reason we should not use default.coffee from a volume is because it changes the defaults of weaver-server, and whatever uses weaver-server should not be an authority on the defaults (that file should _never_ be edited outside of the weaver-server repo, this bit us in the ass before).

Another issue with maintaining a copy of the default config file here is that when new defaults are introduced in weaver-server, this repo misses them and the server will no longer start. This is prevented by using the default.coffee from the image, and only making changes to local.coffee to override the defaults (or production.coffee in the case of NODE_ENV=production)

gvdent:
(And the reason this local.coffee is actually checked in as opposed to the local.coffee in the projectroot/config/ folder is that that file is for a local environment which we do not know, whereas in this file we do know the environment (namely the entire composistion this file is part of).)

carvilsi:
Thanks for the answer @gvdent

PR #341, opened 2018-09-07
adds propagate delete functionality

mingushead:
I certify that:
- [x] Any functionality additions/changes are documented in the README.md
- [x] Any functionality additions/changes are _also_ documented [here](https://github.com/weaverplatform/weaver-docs/blob/master/pages/developers/reference/weaver-sdk-js.md "Weaver-docs")
  another developer
  forget
  target branch
- [x] This is a quality PR


Changelist.md
```diff
@@ -1,5 +1,8 @@
 # Changelist
 
+## develop
+- Adds propogate-delete functionality
```
gvdent: Spelling: `propagates`

test/WeaverNode.test.coffee
```diff
@@ -973,6 +973,35 @@ describe 'WeaverNode test', ->
...
+  it 'should be able to propogate destroy to specified relations', ->
+    a = new Weaver.Node('Grandmother')
+    b = new Weaver.Node('Mother')
+    c = new Weaver.Node('Child')
+    a.relation('gaveBirthTo').add(b)
+    b.relation('gaveBirthTo').add(c)
+
+    a.save().then(->
+      a.destroy(weaver.currentProject().projectId, true, ['gaveBirthTo'])
+    )
```
gvdent: This needs a check that the Child has been destroyed.

mingushead: Added a separate test to test this

bastbijl:
One big issue, this propagetes mechanism is recursive without end.

mingushead:
Bas: _One big issue, this propagates mechanism is recursive without end._

Yeah, that's true, and intentional, and documented. (ps it's not without end, it will stop at any point at which a node doesn't have any relations which match something from the `propagates` arg which was passed during `destroy`

src/WeaverNode.coffee
```diff
@@ -409,7 +409,12 @@ class WeaverNode
     )
 
   # Removes node, with the option to remove it unrecoverable
-  destroy: (project, unrecoverableRemove = false) ->
+  destroy: (project, unrecoverableRemove = false, propogates = []) =>
```
bastbijl: i want to block releasing so i repeat my concern again (sorry for not knowing how to use github properly)

mingushead: this thing about propogates having no end?
I don;t get this, I mean the functionality is intended to be recursive, if the pattern matches then it should have no end until the pattern is not matched any longer..

bastbijl: At some project we use the reference key between different types of nodes with different meanings. The unbounted repeation will result in possibly to many removals.

bastbijl: That's an example. But more simply if you have `a`-`x`->`b`-`x`->`c` and you only want the first relation to be deleted but not the second, this is something that can not be done.

mingushead: so how about if I add a fourth `propagationDepth` argument? Then we're all good, right?

mingushead: Also, permission to allow passing `-1` to `propagationDepth` to allow endless propagation?

bastbijl: I guess a depth argument is ok, because i don't know a better way that is more or less as easy to use as this proposal. But I do not like the -1 option. At least that should not be default. Default should be 1.

mingushead: Default is indeed `1`.
`-1` must be specified by the developer.

mingushead:
changes applied

PR #364, opened 2018-11-22
Using the SDK in React Native

malamili:
This is a WIP, please dont review.

I certify that:
- [x] Any functionality additions/changes are documented in the README.md
- [x] Any functionality additions/changes are _also_ documented [here](https://github.com/weaverplatform/weaver-docs/blob/master/pages/developers/reference/weaver-sdk-js.md "Weaver-docs")
  another developer
  forget
  target branch
- [x] This is a quality PR


package.json
```diff
@@ -11,22 +11,24 @@
...
+    "react-native.js"
+  ],
   "license": "GPL-3.0",
   "repository": {},
   "dependencies": {
     "bluebird": "~3.4.1",
     "circular-json": "^0.3.1",
     "config": "^1.24.0",
-    "cuid": "~1.3.8",
+    "cuid": "2.1.4",
```
malamili: Older cuid didnt work with react native

package.json
```diff
@@ -11,22 +11,24 @@
...
     "localforage": "^1.5.0",
     "lodash": "^4.17.4",
     "lokijs": "^1.4.2",
     "moment": "^2.22.2",
     "pubsub-js": "^1.5.7",
-    "request": "^2.82.0",
```
malamili: this wasnt used at all

package.json
```diff
@@ -11,22 +11,24 @@
...
     "semver": "^5.3.0",
-    "socket.io-client": "^2.0.1",
+    "socket.io-client": "^2.1.0",
```
malamili: just updating, no real reason

package.json
```diff
@@ -35,7 +37,7 @@
     "chai-as-promised": "^6.0.0",
     "coffee-coverage": "2.0.1",
     "coffeeify": "^3.0.1",
-    "coffeescript": "^2.2.4",
+    "coffeescript": "^2.3.2",
```
malamili: just updating, no real reason

src/CoreManager.coffee
```diff
@@ -110,14 +107,14 @@ class CoreManager
...
-      (new Weaver.Plugin(p) for p in plugins)
+      (new PluginConstructor(p) for p in plugins)
     )
 
-  getPlugin: (name) ->
+  getPlugin: (name, PluginConstructor) ->
     @POST("plugin.read", {name}).then((plugin) ->
-      new Weaver.Plugin(plugin)
+      new PluginConstructor(plugin)
     )
```
malamili: Weaver.Plugin might not be set when using react native, thats why we pass the constructor

src/SocketController.coffee
```diff
@@ -1,9 +1,8 @@
 # Libs
-io       = require('socket.io-client')
+io       = require('socket.io-client/dist/socket.io')
```
malamili: the package json isnt correct so react native had trouble finding this module, this works both ways

src/SocketController.coffee
```diff
@@ -16,9 +14,12 @@ class SocketController
     @options.reconnection = true
     @options.query = "sdkVersion=#{pjson.version}&requiredServerVersion=#{pjson.com_weaverplatform.requiredServerVersion}&requiredConnectorVersion=#{pjson.com_weaverplatform.requiredConnectorVersion}"
 
+  requireIO: ->
+    require('socket.io-client/dist/socket.io')
+
```
malamili: In react native, this is the only way it works for requiring socket.io, something wrong with the package json orso for not finding the main file

src/SocketControllerWithStream.coffee
```diff
@@ -0,0 +1,25 @@
...
+  getSocket: (body) ->
+    if body.type isnt 'STREAM'
+      @io
+    else
+      ss(@io)
+
+  STREAM: (path, body) ->
+    @emit(path, body)
+
+module.exports = SocketControllerWithStream
```
malamili: This class has all the extra's that work in a node env, but not in react native

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "weaver-sdk",
-  "version": "11.0.2",
+  "version": "11.0.3-beta.0",
```
Gielert: IMO this should be a minor version, as it adds completely new stuff.

src/CoreManager.coffee
```diff
@@ -110,14 +107,14 @@ class CoreManager
   shout: (message) ->
     @POST("socket.shout", {message})
 
-  listPlugins: ->
+  listPlugins: (PluginConstructor) ->
```
Gielert: So is it strictly necessary now to pass a constructor? Doesn't this break current functionality?

mingushead: +1 for what Giel said. What was the reason for not doing `(PluginConstructor = Weaver.Plugin)->`, or some other backwards-compatible thing here?

package.json
bastbijl: agree

package.json
```diff
@@ -11,22 +11,24 @@
     "requiredConnectorVersion": "^4.12.0",
     "requiredServerVersion": "^3.13.3"
   },
-  "main": "lib/Weaver.js",
+  "main": "lib/node/Weaver.js",
+  "files": [
+    "react-native.js"
```
bastbijl: interesting construction😲

src/SocketController.coffee
gvdent: Is this also the reason the require has to be inside a function body as opposed to at the top of the file?

src/node/Weaver.coffee
```diff
@@ -0,0 +1,16 @@
...
+    super(opts)
+
+    @File        = Weaver.File
+    @Plugin      = Weaver.Plugin
+    @coreManager = new Weaver.CoreManager(SocketController)
+
+module.exports = Weaver
+
+module.exports.Plugin = require('../WeaverPlugin')
+module.exports.File   = require('../WeaverFile')
```
gvdent: These two lines seem to set static members with the same name as instance members defined in the constructor, is this intentional?

package.json
gvdent: It might even be major, I'm unsure that updating to this SDK "just works" without any adjustment to code.

malamili: It works without adjustment to code. Agree with @Gielert though, at least 11.1.0

package.json
malamili: Set to ^

src/CoreManager.coffee
malamili: It doesnt break functionality because this coremanager listPlugins function is private to Weaver.Plugin (that class calls this function and only that class). In Weaver.Plugin the correct constructor is by default passed.  The reason is Weaver.Plugin is not available in coremanager when using react, that is why we pass it through

src/SocketController.coffee
malamili: No different reason, that is because `SocketControllerWithStream` can override this function, as it seems to be different for node environments in relation to react native

src/SocketControllerWithStream.coffee
```diff
@@ -0,0 +1,25 @@
+# Libs
+SocketController = require('./SocketController')
+ss = require('socket.io-stream')
+
+class SocketControllerWithStream extends SocketController
+
+  requireIO: ->
+    require('socket.io-client')
```
malamili: @gvdent see here the override

src/node/Weaver.coffee
malamili: Yes, it is exactly like all other exposes in WeaverBase..i removed them there and put them here only for node env

malamili:
Added yarn.lock file

src/CoreManager.coffee
Gielert: If it's a private function, shouldn't this be prefixed with `_`? @gvdent might have an opionion about this

malamili: CoreManager in its whole should be private to the outside world, so we can guarantee its function use

package.json
mingushead: I'm a tiny bit scared of this, because I once got burned by: https://github.com/jashkenas/coffeescript/issues/5061
That being said, I guess we should go ahead with `2.3.2`, instead of being locked to `2.2.4` until who knows when, because if your sdk compiled, then I guess we're not using the spread operator.
This was actually only an issue on the sdk-lite, which isn't being maintained anymore, so another reason to update.
One reason not to update is that I'm using `2.2.4` on weaver-web, as the spread operator turns out to be really useful for some bread-and-butter redux stuff, so I don't know if you guys agree that we should have a standard coffee version across all our software, in which case I say not to update.
TL;DR: approved, but I mention this anyway for knowledge-sharing, and so that we all appreciate this breaking-change revert that the coffee guys decided on when moving to `2.3`.

package.json
```diff
@@ -11,22 +11,25 @@
     "requiredConnectorVersion": "^4.12.0",
     "requiredServerVersion": "^3.13.3"
   },
-  "main": "lib/Weaver.js",
+  "main": "lib/node/Weaver.js",
```
Gielert: Well, actually I found the breaking change I guess. Since Weaver is not being put on the window anymore, it breaks for all browsers that import the `weaver-sdk-full.js`. I just noticed that the main file changed, and this file doesn't put Weaver on the window. Or is there some sort of workaround?

malamili: Weird, the line to put it on the window is still there though

